PR #39, opened 2026-05-03
fix: handle embedded docx text newlines

jan-kubica:
JK:
I am on the verge here... comes from real .docx, probably converted from another format to .docx in a slightly broken state and the MS Word seems to be do similar logic. At the other hand we ar epatching noncompliant .docx format, which opens up a tricky category of issues. 

## Summary
- convert embedded CR/LF characters in DOCX text runs into explicit hard-break nodes during Folio import
- apply the same splitting to hyperlink text so layout measurement and browser rendering stay aligned
- add a regression for DOCX text nodes containing mixed newline forms

## Test Plan
- bun --filter @stll/folio typecheck
- bun --filter @stll/folio lint
- bun test packages/folio/src/core/prosemirror/conversion/toProseDoc.test.ts packages/folio/src/core/layout-bridge/toFlowBlocks.test.ts
- pre-push hook: format, i18n, lint, typecheck

---
_Mirror of original PR #1052 by @jan-kubica_

### DIFF
--- a/packages/folio/src/core/prosemirror/conversion/toProseDoc.test.ts
+++ b/packages/folio/src/core/prosemirror/conversion/toProseDoc.test.ts
@@ -98,6 +98,55 @@ describe("toProseDoc", () => {
     expect(commentMark?.attrs.commentId).toBe(42);
   });
 
+  test("converts embedded DOCX text newlines to hard breaks", () => {
+    const document: Document = {
+      package: {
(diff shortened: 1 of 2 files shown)


jan-kubica:
closing, trying to fix broken .docx may ultimately cause more harm than good, accepting that in some edgecases we drift from what MS Word would do & display

PR #41, opened 2026-05-03
feat(ai): require BYOK when no instance keys are provisioned

jan-kubica:
## Summary

Adds a backend gate and frontend UX so users can never reach an AI feature that would 500 because no provider key is available.

- New env flag `REQUIRE_PERSONAL_AI_KEY` lets operators force BYOK even on instances that *do* have provisioned keys, useful for shared deployments without metering where each org's costs land on their own provider account.
- `read-ai-config` now returns `instanceProvisioned: boolean` so the frontend knows whether AI is usable without org BYOK.
- `requireAIAvailable(orgConfig)` Result helper, applied at every AI handler entry: chat send-message, search refine + summary, case-law analysis generate, entities organize-suggestions. Returns 403 to clients when neither org BYOK nor instance keys are available.
- `<RequireAIKey>` route-wraps `/chat`, the most prominent AI surface. Shows an "AI key required" card linking to `/settings/organization/ai`.
- `useAIAvailable()` hook exported alongside for component-level gating in shared dialogs (search summary buttons, file organizer suggestions). Wiring those is left to follow-up PRs.
- i18n keys `ai.keyRequired.{title,description,cta}` added to all 12 languages. English is the source; non-English values are placeholders pending Lingo or manual translation.

## Behavior

`instanceProvisioned = false` IFF `REQUIRE_PERSONAL_AI_KEY=true` OR (no env keys for any provider). When false AND no org BYOK → user sees the gate.

## Test plan

- [ ] With `REQUIRE_PERSONAL_AI_KEY=false` and at least one provider env key set, /chat works normally (no gate)
- [ ] With `REQUIRE_PERSONAL_AI_KEY=true` and no org BYOK, /chat shows the gate card with link to settings
- [ ] With `REQUIRE_PERSONAL_AI_KEY=true` and an org BYOK saved at /settings/organization/ai, /chat works
- [ ] When the gate is active, calling `POST /chat`, `POST /search/summary`, `POST /search/refine`, `POST /case/decisions/:id/analysis`, `POST /entities/:wsId/organize-suggestions` directly returns HTTP 403
- [ ] Background jobs (case-law citation polarity LLM fallback) keep working — they have their own fallback path and are intentionally not gated

## Follow-ups (out of scope)

- Apply `useAIAvailable` to search-dialog AI buttons, file-organizer "Generate suggestions" button, case-decision analysis sidebar to hide/disable AI affordances inline
- Translate the new i18n keys into the 11 non-English languages (Lingo bot was killed during the repo migration; these need manual or replacement-tool translation)

### DIFF
--- a/apps/api/src/env.ts
+++ b/apps/api/src/env.ts
@@ -47,6 +47,16 @@ const envApi = createEnv({
     ANTHROPIC_API_KEY: v.optional(v.string()),
     GOOGLE_AI_API_KEY_EU: v.optional(v.string()),
     GOOGLE_AI_API_KEY_CH: v.optional(v.string()),
+    /**
+     * Force orgs to supply their own AI key (BYOK) even if the
+     * instance has provisioned provider keys. Useful for shared
+     * deployments without metering where the operator wants
+     * costs to land on each org's own provider account.
+     */
+    REQUIRE_PERSONAL_AI_KEY: v.optional(
+      v.pipe(v.string(), v.parseBoolean()),
+      "false",
+    ),
     REDIS_URL: v.pipe(v.string(), v.url()),
     USE_MOCK_AI: v.optional(v.pipe(v.string(), v.parseBoolean()), "false"),
     BETTER_AUTH_SECRET: v.pipe(v.string(), v.minLength(32)),
--- a/apps/api/src/handlers/case-law/analysis/generate.ts
+++ b/apps/api/src/handlers/case-law/analysis/generate.ts
@@ -36,6 +36,7 @@ import {
   getModelForRole,
   getModelInfoForRole,
   getTemperatureForRole,
+  requireAIAvailable,
 } from "@/api/lib/ai-models";
 import type { OrgAIConfig } from "@/api/lib/ai-models";
 import { captureError } from "@/api/lib/analytics";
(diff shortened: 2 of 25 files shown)


jan-kubica:
Addressed all four bot-review comments in 9e37e6055:

- **Gemini (high)** — `hasInstanceProvider` now includes `USE_MOCK_AI`, `GOOGLE_AI_API_KEY_EU`, `GOOGLE_AI_API_KEY_CH`
- **Codex P1** — explicit `AI_PROVIDER` paths validate matching credentials (e.g. `AI_PROVIDER=openai` requires `OPENAI_API_KEY`); no more false 'available' state
- **Codex P2 (gate bypass for non-admin)** — new `/organization-settings/ai-availability` endpoint scoped to `workspace:read` so any org member can read availability without admin perm; `RequireAIKey`/`useAIAvailable` now consume it and treat query errors as unavailable
- **Codex P2 (case-law 200 vs 403)** — gate moved into the route handler (`requireAIAvailable` direct `yield*`) so the endpoint returns HTTP 403 consistently with the other AI handlers

jan-kubica:
Gemini, the latest 3 findings are all false positives — they ask about pre-existing code that isn't in the PR diff. Verified locally:

- `status` function is imported on line 6 of `apps/api/src/handlers/search/ai.ts`: `import { status, t } from "elysia";` — pre-existing, used by every handler in that file
- `AI_PROVIDER_BASE_URL` is defined on line 28 of `apps/api/src/env.ts` in the `envApi` schema with `v.optional(v.pipe(v.string(), v.url()))` — pre-existing

Both are tracked in main, just not part of this PR's diff. The full files compile and pass typecheck.

PR #97, opened 2026-05-04
chore: drop apple icon for consistency

jan-kubica:


jan-kubica:
actual fix is removin both icons, just noise..

PR #112, opened 2026-05-04
feat: update landing showcase screenshots

wzarek:


jan-kubica:
does not look clean enough i would rather do it properly with component-like style

PR #124, opened 2026-05-04
fix: improve ask-user keyboard flow

jan-kubica:
## Summary
- let option answers advance focus to the next ask-user question
- let Enter and forward Tab confirm text answers and advance through the ask-user card
- submit from the final question when the keyboard flow reaches the end

## Test plan
- bun --filter @stll/web lint
- bun --filter @stll/web typecheck


jan-kubica:
ultimately found annoying

PR #140, opened 2026-05-04
fix: clean up ai settings copy

jan-kubica:
## Summary
- remove the AI data-region explanatory string and hide that help text for supported providers
- improve AI key and AI settings copy across locales
- regenerate i18n message types

## Test plan
- bun run lint
- bun run format
- bun run typecheck
- bun run test
- bun run i18n:check


jan-kubica:
supersededd

PR #141, opened 2026-05-04
feat(ci): synthetic prod check every 5 min

jan-kubica:
Cron workflow that hits live prod every 5 minutes and verifies user-critical paths still respond. On success pings a heartbeat URL; on any failure the heartbeat is skipped and the monitoring service pages once it times out.

## Checks (v1)
- API `/health` returns ok
- Frontend / returns the SPA shell (asserts the bootstrap div, not just any 200 — SPA fallback returns 200 for everything)
- API auth endpoint OPTIONS preflight returns correct CORS allow-origin (catches the silent class of bugs where browser requests fail with no backend log)
- Desktop updater manifest returns valid JSON with a `.version` field

## Setup
Add a repo secret `SYNTHETIC_HEARTBEAT_URL` with the heartbeat URL from your monitoring service. Without it the workflow still runs and reports failures via the GitHub Actions UI but won't actively page.

### DIFF
--- a/.github/workflows/synthetic-prod-check.yml
+++ b/.github/workflows/synthetic-prod-check.yml
@@ -1,0 +1,107 @@
+name: Synthetic Prod Check
+
+# Runs every 5 minutes against the live app, exercises the user-
+# critical paths, and pings the heartbeat monitor on success. Skips
+# the ping on failure so the monitor times out and pages.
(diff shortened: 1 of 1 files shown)


jan-kubica:
PoC but directionally wrong

PR #189, opened 2026-05-05
fix: table property and grid UI

nnad3N:
## Proposed change
creating new property now only works on the header col, because it was clashing with expending the row

image
image
image

## Checklist

- [ ] BUILD ASSURANCE | Code builds correctly and without any errors or warnings.
- [ ] TESTING | Changes tested, including in different conditions (dark mode, language, narrow screen, etc).
- [ ] DARK MODE SUPPORT | User can use the webapp in dark mode. Make sure your changes are visible in both light and dark mode.
- [ ] ISSUE LINKED | Link an issue to this PR (not by mentioning it in the description, but by using the GitHub issue linking feature)


jan-kubica (changes requested):
I think we should aim visually for a "full table view", even if just part of the table is populated, with the + rows being on the very bottom and far right, no matter the size of table. Cf i you open google sheets. Also cf v7 design.

jan-kubica:
given the complexity, opening followup and merging this as an improvement of current state https://github.com/stella/stella/issues/197

PR #226, opened 2026-05-06
chore: clean startup scripts

jan-kubica:
chore: clean startup scripts


jan-kubica:
pushing back, this would pollute the code going forward imo

PR #229, opened 2026-05-06
chore: relicense from AGPL-3.0 to Apache-2.0

jan-kubica:


jan-kubica:
sbom is autogenerated

PR #261, opened 2026-05-09
fix: expose build commit metadata

jan-kubica:
Expose the build commit alongside the app version in the API health response and app shell.


jan-kubica:
/codex:review

CC on behalf of @sok0

PR #271, opened 2026-05-09
feat(folio): port upstream eigenpal HF unification + follow-ups

### LINKED ISSUE #91: bug(folio): investigate Footer and Header writing
### Reproduction

investigate wirting and changing text in the header and footer

### Expected behavior

... it works

### LINKED ISSUE #130: bug(folio): overlapping text
### Reproduction

Image

[NVCA Series A - Stock Purchase Agreement.docx](https://github.com/user-attachments/files/27370244/NVCA.Series.A.-.Stock.Purchase.Agreement.docx)

### Expected behavior

render as in antive word

jan-kubica:
## Summary

- Routes header/footer content through the same ProseMirror -> FlowBlock -> measureBlocks chain as the body. New shared `core/layout-bridge/headerFooterLayout.ts` module replaces the legacy paragraph-only converter in `PagedEditor.tsx` (-637 lines of shadow-stack code).
- Adds the table arm to `renderHeaderFooterContent` so HF tables stop being silently dropped, plus the floating-table position resolver and `RenderContext.positioning` field.
- Closes upstream #356/#357/#358 (HF tables in paginated view + shared interaction model + measurement parity) and the #379-#382 follow-ups (positioning context, style-inherited spacing strip, trailing empty-paragraph-after-table suppression, floating tables in HF). `convertTable` theme threading for themeColor cell shading remains orthogonal upstream divergence.

CC on behalf of @jan-kubica.

## Test plan

- [x] Folio typecheck clean.
- [x] Folio lint clean.
- [x] 571/571 folio unit tests pass.
- [x] Web typecheck clean (consumer).
- [ ] Manual playground verification on a DOCX with HF tables (regular + floating), HF anchored images, NVCA Series A SPA template (multi-section + first-page header + paginated footnotes).
- [ ] Visual / Playwright snapshot suite.

### DIFF
--- a/packages/folio/src/core/layout-bridge/headerFooterLayout.test.ts
+++ b/packages/folio/src/core/layout-bridge/headerFooterLayout.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, test } from "bun:test";
+
+import type { FlowBlock, Measure } from "../layout-engine/types";
+import type { HeaderFooterMetrics } from "./headerFooterLayout";
+import { calculateHeaderFooterVisualBounds } from "./headerFooterLayout";
+
+const metrics: HeaderFooterMetrics = {
+  section: "header",
+  pageSize: { w: 600, h: 800 },
+  margins: {
+    top: 100,
+    right: 72,
+    bottom: 100,
+    left: 72,
+    header: 48,
+    footer: 48,
+  },
+};
+
+const tableMeasure = (totalHeight: number): Measure => ({
+  kind: "table",
+  rows: [],
+  columnWidths: [],
+  totalWidth: 120,
+  totalHeight,
+});
+
+describe("calculateHeaderFooterVisualBounds", () => {
+  test("accounts for page-anchored floating table bounds", () => {
+    const blocks: FlowBlock[] = [
+      {
+        kind: "table",
+        id: "floating-table",
+        rows: [],
+        floating: {
+          vertAnchor: "page",
+          tblpYSpec: "bottom",
+        },
+      },
+    ];
+
+    const bounds = calculateHeaderFooterVisualBounds(
+      blocks,
+      [tableMeasure(50)],
+      0,
+      metrics,
+    );
+
+    expect(bounds).toEqual({ visualTop: 0, visualBottom: 752 });
+  });
+
+  test("defaults floating table bounds to the source cursor position", () => {
+    const blocks: FlowBlock[] = [
+      {
+        kind: "paragraph",
+        id: "intro",
+        runs: [{ kind: "text", text: "Intro" }],
+      },
+      {
+        kind: "table",
+        id: "floating-table",
+        rows: [],
+        floating: {},
+      },
+    ];
+
+    const bounds = calculateHeaderFooterVisualBounds(
+      blocks,
+      [
+        {
+          kind: "paragraph",
+          lines: [],
+          totalHeight: 20,
+        },
+        tableMeasure(50),
+      ],
+      20,
+      metrics,
+    );
+
+    expect(bounds).toEqual({ visualTop: 0, visualBottom: 70 });
+  });
+});
--- a/packages/folio/src/core/layout-bridge/headerFooterLayout.ts
+++ b/packages/folio/src/core/layout-bridge/headerFooterLayout.ts
@@ -281,7 +281,18 @@ export function calculateHeaderFooterVisualBounds(
       visualBottom = Math.max(visualBottom, paragraphBottomY);
 
       for (const run of block.runs) {
-        if (run.kind !== "image" || !run.position) {
+        if (run.kind !== "image") {
+          continue;
+        }
+        // Match the extraction/normalization classification: any image
+        // floating by `position` OR by `wrapType`/`displayMode` is
+        // rendered absolutely and stripped from paragraph measurement,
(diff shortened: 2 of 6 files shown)


jan-kubica:
Addressed two more bot points:

- **`spaceBefore` swallowed** (Codex P2 #3): paragraph fragment top is now `cursorY + spaceBefore`. The body's layout engine already does this in `addFragment`; HF needs the same offset because `renderParagraphFragment` paints the first line at fragment top, so without it `paragraphMeasure.totalHeight` reserves the spaceBefore but never visually applies it.
- **Floating tables in visualBounds** (Codex P2 #4): `calculateHeaderFooterVisualBounds` now expands bounds at the floating table's inline `cursorY` position (conservative). Exact `(tblpX, tblpY)`-resolved bounds would duplicate `resolveHeaderFooterFloatingTablePosition`; for tables anchored far from the in-flow stream the bound undercounts. Without this, a floating-table-only HF had `height = 0` and the table was lost from the body push-down extender entirely.

CC on behalf of @jan-kubica.

jan-kubica:
One more bot point addressed:

- **Stripped inherited HF spaceBefore** (Codex P2 #5): the spaceBefore offset now gates on `spacingExplicit.before`. `normalizeHeaderFooterMeasureBlocks` strips style-inherited `spaceBefore` from the measurement copy (#380), so `totalHeight` excludes it; reading the original block's `spacing.before` for the render offset would have shifted the line below its reserved space and desynced `cursorY` from `totalHeight` on subsequent paragraphs.

CC on behalf of @jan-kubica.

jan-kubica:
One more bot point addressed:

- **Footer flow origin vs. visual extent mismatch** (Codex P2 #6): when `visualBottom > flowHeight` (newly possible after my floating-table visual-bounds fix), the footer container was positioned at `containerTop = page.h - footerDistance - actualFooterHeight` while the resolver's `flowTop` was computed from `flowHeight`. The two coord origins diverged by the overflow amount, so page/margin-anchored floating tables in footers rendered too high and in-flow content shifted up with them. Anchored the footer container at the natural flow origin (`page.h - footerDistance - flowHeight + visualTop`) and let `actualFooterHeight` extend the container downward into the bottom margin. In-flow content stays at the natural footer line; below-flow overflow paints into the bottom margin (the price of unwrapped overflow). Resolver's `flowTop` now matches where in-flow content actually paints.

CC on behalf of @jan-kubica.

jan-kubica:
Two more bot points addressed:

- **Inline HF table alignment** (Codex P2 #7): inline HF tables now compute `left` from `block.justification` (`center`/`right`) and `block.indent`, mirroring the body pagination's `desiredX` logic. Previously the arm pinned `left: 0`, so centered or indented HF tables rendered flush-left.
- **HF text box measured width** (Codex P2 #8): synthetic `TextBoxFragment` uses `measure.width` instead of `contentWidth`. `measureBlocks` computes `measure.width` and caches `innerMeasures` against the authored text box width; passing the full HF `contentWidth` made the outer box the wrong size and forced inner re-wrap inconsistent with the cached measures.

CC on behalf of @jan-kubica.

jan-kubica:
Bot point addressed:

- **`inside`/`outside` floating-table specs** (Codex P2): `resolveHeaderFooterFloatingTablePosition` now aliases `tblpXSpec="inside"` to left and `"outside"` to right, matching the body layout path. Without this, an outside-anchored floating table in HF rendered at left = 0 instead of the right edge.

CC on behalf of @jan-kubica.

jan-kubica:
Bot point addressed:

- **Suppress only the canonical trailing empty paragraph** (Codex P2): `normalizeHeaderFooterMeasureBlocks` now restricts `suppressEmptyParagraphHeight` to the LAST block only (when it's an empty paragraph after a table). HF content shaped `[table, blank, paragraph, blank, table, blank]` previously had every blank-after-table collapsed; only the final blank is the canonical OOXML trailing element, so the inner blanks now keep their authored vertical space.

CC on behalf of @jan-kubica.

jan-kubica:
Bot point addressed:

- **Lift HF floating images by wrapType, not just position** (Codex P2): match the body's `isFloatingImageRun` classification both when extracting floating images from HF paragraphs (`renderHeaderFooterContent`) and when stripping them from the measurement copy (`isAnchoredImageRun` in `headerFooterLayout`). After the unified pipeline, HF paragraphs receive the same `toFlowBlocks` image runs as the body — square/tight/through/behind/inFront wrap types and `displayMode: "float"` runs may not carry explicit `<wp:positionH>`/`<wp:positionV>`. The inline renderer skips them; without the matching extraction, they were silently dropped from the rendered header. Synthesize an empty `position` for the position-less case; the float helpers fall through to a paragraph-relative default, mirroring the body path.

CC on behalf of @jan-kubica.

jan-kubica:
Bot point addressed:

- **Wrap-type-only HF images in visual bounds** (Codex P2): `calculateHeaderFooterVisualBounds` now uses the same broader classification as the extraction and normalization paths (`run.position` OR `isFloatingImageRun(run)`). The previous `run.position`-only check skipped wrap-type-only images, so a header containing such an image had `visualBottom` equal to the stripped paragraph height — `computeHeaderFooterMarginExtender` then under-reserved body push-down and the image overlapped body text even though it rendered correctly.

CC on behalf of @jan-kubica.

PR #280, opened 2026-05-10
feat(folio): port upstream eigenpal TIFF -> PNG image conversion

jan-kubica:
## Summary

- Ports eigenpal/docx-editor #146 (commits 0c4486718 + 41f923de4): browsers don't render TIFF in \`\`, so DOCX-embedded TIFFs are decoded with utif2 and re-encoded as PNG via the Canvas API at parse time. The mediaFile's mimeType flips to image/png so consumers see the effective format and re-export writes PNG instead of the raw TIFF.
- Falls back to the lazy raw data URL when Canvas is unavailable (headless parsing) so round-trips still survive without an in-browser preview.

Originally part of an "upstream branch sync" bundle covering four fix branches; the other three (\`feat/column-separator-line\`, \`fix/drag-select-autoscroll\`, \`fix/firefox-right-click-selection\`) were already in folio from earlier syncs, so only TIFF support actually needs porting.

CC on behalf of @jan-kubica.

## Test plan

- [x] Folio typecheck clean.
- [x] Folio lint clean.
- [x] 571/571 folio unit tests pass.
- [ ] Manual playground verification on a DOCX containing a TIFF embedded image (e.g. eigenpal's \`tiff-test.docx\` fixture).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -328,6 +328,7 @@
         "prosemirror-tables": "^1.8.5",
         "prosemirror-transform": "^1.12.0",
         "prosemirror-view": "^1.41.8",
+        "utif2": "^4.1.0",
       },
       "devDependencies": {
         "@playwright/test": "^1.59.1",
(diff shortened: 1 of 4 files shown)


jan-kubica:
All three bot points addressed:

- **Data/mimeType skew** (Gemini high): `convertTiffToPngDataUrl` now returns `{ dataUrl, data }` so the decoded PNG bytes flow back into `MediaFile.data`. Previously the file was labelled `image/png` but the buffer still held TIFF, so re-export wrote a mislabelled file. Filename `.tiff`/`.tif` now also flips to `.png`.
- **Eager pixel allocation** (Codex P1): added a 64 MP cap checked from the IFD's declared `width`/`height` BEFORE `UTIF.toRGBA8`. A crafted TIFF whose ZIP-compressed bytes fit under `unzipDocx`'s media cap can still declare arbitrary dimensions; without this, `toRGBA8` allocates the full width × height × 4 RGBA buffer and can hang or crash the tab.
- **Case-insensitive mime check** (Gemini medium): `isTiffMimeType` now lowercases before comparing.

CC on behalf of @jan-kubica.

jan-kubica:
Bot point addressed:

- **TIFF dimensions read from IFD tags** (Codex P2): `UTIF.decode()` returns IFDs whose `width`/`height` properties are populated only by `UTIF.decodeImage()`. My pre-decode safety check was therefore reading `undefined` and bailing out for every TIFF, so no DOCX TIFF actually reached the conversion path. Now reads dimensions from the raw IFD tags (`t256` ImageWidth, `t257` ImageLength) for the pixel cap, then proceeds to `decodeImage` for the actual conversion.

CC on behalf of @jan-kubica.

jan-kubica:
One bot point addressed, one deferred to follow-up:

- **Canvas check before TIFF decode** (Codex P2): moved the DOM availability check to the top of `convertTiffToPngDataUrl`. Headless parses (no `document`) previously ran `decodeImage` + `toRGBA8` for the full pixel buffer (up to ~256 MB at the 64 MP cap) before noticing Canvas was missing. Bailing first costs nothing when the conversion can't succeed anyway.
- **Re-export still preserves original TIFF** (Codex P2 #2): real bug, but the fix is significantly bigger than this PR — `repackDocx` copies media verbatim from the original ZIP, and `processNewImages` only handles newly-inserted images. Threading converted-TIFF bytes through requires patching the rels Target, the media part, and `[Content_Types].xml` for each converted file. I'll file this as a follow-up issue rather than ballooning this PR; the user-visible win here (TIFFs now render in the in-browser preview) is delivered, and round-tripping a TIFF still preserves it as TIFF (a downgrade only relative to the upstream commit's promise, not relative to current main).

CC on behalf of @jan-kubica.

PR #313, opened 2026-05-12
chore(ci): wave 1 supply-chain hardening

jan-kubica:
## Summary

Wave 1 of the CI hardening pass — low-risk, no behavioural changes. Larger items (cache scoping on release, environment gates, OIDC-only npm publish) will follow in separate PRs.

- `permissions: {}` at the root of 4 workflows (`ci.yml`, `db-migrations.yml`, `nightly-typecheck.yml`, `audit-branch-protection.yml`) — every job already declares the perms it actually needs, so this is purely a default-deny safeguard against future drift.
- Drop `secrets: inherit` from `pr-lint.yml`. The reusable workflow at `stella/.github/.github/workflows/pr-lint.yml@9aba86f1` only consumes the auto-provided `GITHUB_TOKEN`, so inheritance was a no-op. Removing it means any future secret usage by the reusable becomes a reviewable diff at the caller.
- Document the `pull_request_target` trust-boundary invariant in `cla.yml` — checking out PR head code from this workflow is the attack shape behind tj-actions / Ultralytics / Nx. Adding a header comment so the rule is visible to future contributors.
- Add `apps/desktop/src-tauri/Cargo.lock` to `CODEOWNERS`. The Rust lockfile feeds the signed desktop build; it should sit behind the same review gate as `Cargo.toml`.

## Test plan

- [x] `actionlint` clean on all six modified workflows
- [ ] CI green on this PR (this PR exercises the new `ci.yml` and `pr-lint.yml` because `pull_request` events use the workflow file from the PR branch)
- [ ] `pr-lint` check passes — confirms the reusable workflow still works without `secrets: inherit`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,6 +13,7 @@ apps/desktop/src-tauri/Cargo.toml       @jan-kubica
 apps/desktop/src-tauri/Cargo.lock       @jan-kubica
 apps/desktop/src-tauri/build.rs         @jan-kubica
 apps/desktop/src-tauri/tauri.conf.json  @jan-kubica
+apps/desktop/src-tauri/capabilities/    @jan-kubica
 
 # Privileged bridge — desktop code with elevated OS access.
 apps/desktop/src-tauri/src/             @jan-kubica
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,6 +10,7 @@ scripts/                                @jan-kubica
 # Desktop signing surface — build runs with signing keys live.
 apps/desktop/package.json               @jan-kubica
 apps/desktop/src-tauri/Cargo.toml       @jan-kubica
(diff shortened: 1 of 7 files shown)


jan-kubica:
CC on behalf of @jan-kubica

Both bot findings addressed in b9eb4a9a7:

- **codex (audit-branch-protection.yml):** correct — `permissions: {}` at workflow root with no per-job override would have capped the called reusable's `GITHUB_TOKEN` to no scopes via the caller-intersection rule, breaking its `actions/checkout`. Added `permissions: contents: read` on the `audit` job to restore the reusable's declared minimum.
- **gemini (CODEOWNERS):** correct — `apps/desktop/src-tauri/capabilities/` defines the Tauri v2 permission boundary for the renderer, so it belongs alongside `tauri.conf.json` under the same review gate.

The same intersection concern does not affect `ci.yml`, `db-migrations.yml`, or `nightly-typecheck.yml` because their jobs all declare their own `permissions:` blocks.

PR #317, opened 2026-05-13
feat(api): accept DB connection components alongside DATABASE_URL

jan-kubica:
Add `resolveDatabaseUrl()` so the API and migrations accept either a full `DATABASE_URL` or a set of components (`DB_HOST`, `DB_PORT`, `DB_USER`, `DB_PASSWORD`, `DB_NAME`, optional `DB_SSLMODE`). Lets the password come from a different source than the rest of the connection metadata.

- `DATABASE_URL` flow is unchanged.
- Components: user/password/host/port/name are percent-encoded; `DB_SSLMODE` is restricted to TLS-enforcing modes (`require`, `verify-ca`, `verify-full`) and defaults to `require`.

## Test plan
- [x] `bun --filter @stll/api test src/db-url.test.ts`
- [x] `bun --filter @stll/api typecheck` and lint
- [x] Local API boot with existing `DATABASE_URL` unchanged

### DIFF
--- a/apps/api/drizzle.config.ts
+++ b/apps/api/drizzle.config.ts
@@ -1,8 +1,10 @@
 import { defineConfig } from "drizzle-kit";
 
-// Fallback keeps knip and CI happy when DATABASE_URL is unset.
+import { resolveDatabaseUrl } from "./src/db-url";
+
+// Fallback keeps knip and CI happy when no DB env is set.
 // drizzle-kit push/generate will fail fast on connection, not silently.
-const url = process.env.DATABASE_URL ?? "postgresql://invalid";
+const url = resolveDatabaseUrl() ?? "postgresql://invalid";
 
 export default defineConfig({
   out: "./drizzle",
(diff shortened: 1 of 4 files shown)


jan-kubica:
CC on behalf of @jan-kubica

Thanks @gemini-code-assist — all three points addressed in fe0e2fe1f:

1. Empty `DB_PASSWORD` is now allowed (`=== undefined` check). Threw `DB_PASSWORD` only when truly unset, not on empty string.
2. Dropped `encodeURIComponent` on `DB_HOST` and `DB_PORT` so bracketed IPv6 literals like `[::1]` survive intact. Encoding stays on the userinfo and path-segment components, where escape-into-other-component is the real risk.
3. Renamed the test to `percent-encodes the database name (path segment)` to reflect what it actually covers, plus added explicit tests for IPv6 host and empty password.

jan-kubica:
CC on behalf of @jan-kubica

Thanks @chatgpt-codex-connector — addressed in a46f40b3e. Validating `DB_HOST` for URL delimiters (`/ ? # @` and whitespace) and `DB_PORT` for numeric-only at the boundary, while still leaving them unencoded so IPv6 literals like `[::1]` survive. Added 9 new tests for the rejection cases.

PR #330, opened 2026-05-14
refactor(oxlint-plugins): extract shared AST helpers

jan-kubica:
Each custom rule in `.oxlint-plugins/` independently re-implemented Identifier/Literal narrowing, property-name extraction, and dot-notation callee walking. Six rule files each repeated the `node.type === "Identifier"` shape check.

Adds `.oxlint-plugins/utils.ts` with `isIdentifier`, `isStringLiteral`, `getPropertyName`, `isMemberAccess`, `isCallTo`, `getCalleeName`, `unwrapExpression`, `getImportedName`. All loosely typed (`unknown` → type guard) so any rule file can adopt them without per-call ceremony.

Migrates three rules as proof of fit:
- `no-body-ownership-ids`: drops inline Identifier-narrow + key-extraction logic.
- `no-raw-user-id-schema`: drops local `getPropertyName` and `getCalleeName`.
- `require-safe-route-handlers`: drops local `isIdentifier`, `isStringLiteral`, `getPropertyName`.

Net `-50` lines across the rule files. Remaining 14 rules can adopt the helpers as they're touched.

## Test plan
- [x] `bun --bun oxlint -c oxlint.config.ts --type-aware` clean on `apps/api`, `apps/web`, `packages/ui`.
- [x] Smoke-tested all three refactored rules with positive and negative cases against synthetic handler/route files; positive cases trigger with the original messages, negative cases do not.
- [x] `oxfmt --check` clean on touched files.

### DIFF
--- a/.oxlint-plugins/mcp-security.ts
+++ b/.oxlint-plugins/mcp-security.ts
@@ -9,30 +9,15 @@
 //   2. OAuth client joins must stay behind the typed chat-time MCP connection
 //      loader, which normalizes raw nullable DB rows into a discriminated union.
 
+import { getPropertyName, isCallTo, isIdentifier } from "./utils.ts";
+
 const MCP_OAUTH_CLIENTS = "mcpOAuthClients";
 const OAUTH_CLIENT_JOIN_ALLOWED_FILES = [
   "apps/api/src/handlers/chat/tools/external-mcp-tools.ts",
 ];
 
-const getPropertyName = (node) => {
-  if (!node) {
-    return null;
-  }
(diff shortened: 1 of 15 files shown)


jan-kubica:
CC on behalf of @jan-kubica: the implementation matches the behaviour of the original local helpers in `no-raw-user-id-schema` and `no-unbranded-ownership-id-param`; both used `Set.has(name)` checks that accepted the property-only form. Updated the docstring to state this explicitly and flag the consideration for callers using fixed allowlists. Fix in 27dc67eec.

PR #345, opened 2026-05-15
chore(folio): extract dialog tree into DocxEditorDialogs

jan-kubica:
## Summary
- Move the six `React.lazy()` dialog imports and the ~110-line dialog render block from `DocxEditor.tsx` into a new `DocxEditorDialogs.tsx`. Editor body shrinks by ~140 lines (still 4,196 — many more PRs to go).
- The new component accepts one `Mount` object per dialog mirroring that dialog's own prop shape. State and apply handlers continue to live in `DocxEditor.tsx` and get passed down; no new state model.
- Promote inline `import("./dialogs/ImagePropertiesDialog").ImagePropertiesData` cast to a real import; pull the IIFE that built `currentData` for the image properties dialog into a `buildImagePropertiesData` helper.

Stacks on the `DocxEditor.tsx` decomposition work; depends on no prior PR.

## Test plan

### DIFF
--- a/packages/folio/src/components/DocxEditor.tsx
+++ b/packages/folio/src/components/DocxEditor.tsx
@@ -18,8 +18,6 @@ import {
   useMemo,
   forwardRef,
   useImperativeHandle,
-  lazy,
-  Suspense,
 } from "react";
 import type { CSSProperties } from "react";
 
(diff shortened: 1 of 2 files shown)


jan-kubica:
Re Gemini suggestions:

- The four "drop the spread pattern" suggestions (lines 141, 158, 168, 181): the spread is required by `exactOptionalPropertyTypes: true`. The dialog props declare `currentProps?: T` (not `T | undefined`), so passing an `undefined` literal is a type error. Confirmed by trying the simplification — `bun --filter @stll/folio typecheck` rejects it.
- Memoizing `imagePropertiesCurrentData`: React Compiler is enabled in the Vite build (CLAUDE.md says prefer plain React over prophylactic `useMemo`/`useCallback`/`React.memo`). The helper is cheap (four nullish checks + one object alloc) and the surrounding `imageProperties` mount object is also re-created each render, so memoizing the inner field doesn't change downstream rerender behavior.

CC on behalf of @jan-kubica

PR #346, opened 2026-05-15
chore(folio): extract zoom and page-indicator state into a hook

jan-kubica:
## Summary
- Move zoom level (with viewport-anchor capture/replay) and the page-indicator overlay (current/total page driven by scroll position) out of `DocxEditor.tsx` and into a new `useZoomAndPageInfo` hook.
- Remove `state.zoom` from the kitchen-sink `EditorState` — the editor now reads `zoom` directly from the hook return.
- Drop the throwaway `handleZoomChange` wrapper (toolbar binds directly to `setZoomWithViewportAnchor`).
- Promote magic numbers (page gap, top padding, fade ms) to named constants inside the hook.

`DocxEditor.tsx` shrinks by 126 lines net. No behavior change.

## Test plan


jan-kubica:
Gemini's two suggestions are addressed in 99e75e0f:

- `DEFAULT_PAGE_GAP` and `VIEWPORT_PADDING_TOP` are now `export`ed from `packages/folio/src/paged-editor/PagedEditor.tsx` and imported by the hook; the duplicated literals are gone.
- The scroll-listener effect now carries a comment explaining the deliberate ref-current capture into a render-scoped const (the container mounts on a later render, then doesn't swap).

CC on behalf of @jan-kubica

PR #347, opened 2026-05-15
chore(folio): extract keyboard shortcuts into a hook

jan-kubica:
## Summary
- Move the document-level keydown effect out of `DocxEditor.tsx` into a new `useKeyboardShortcuts` hook. The hook handles Cmd/Ctrl+F, Cmd/Ctrl+H, Cmd/Ctrl+P, and Delete/Backspace for fully-selected tables.
- `handleDirectPrint` stays in `DocxEditor.tsx` for now (it depends on `containerRef`, `t`, and the local toast); the hook takes it as `onDirectPrint`.
- The whole-table deletion branch goes into a `tryDeleteSelectedTable` helper so the keydown handler reads top-to-bottom instead of being nested two levels deep.

No behavior change.

## Test plan

### DIFF
--- a/packages/folio/src/components/DocxEditor.tsx
+++ b/packages/folio/src/components/DocxEditor.tsx
@@ -199,6 +199,7 @@ import { useFindReplace } from "./hooks/useFindReplace";
 import { useHeaderFooterEditor } from "./hooks/useHeaderFooterEditor";
 import { useHyperlinkHandlers } from "./hooks/useHyperlinkHandlers";
 import { useImageHandlers } from "./hooks/useImageHandlers";
+import { useKeyboardShortcuts } from "./hooks/useKeyboardShortcuts";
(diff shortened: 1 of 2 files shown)


jan-kubica:
Gemini's three suggestions addressed in 77e1a5f3f:

- **High**: Skip Delete/Backspace table-deletion when focus is in an input/textarea/non-editor contenteditable. New `isFocusInInputLike` helper, checked before invoking `tryDeleteSelectedTable`.
- **Medium**: `isMacPlatform` now guards `typeof navigator === "undefined"` and matches iOS variants too.
- **Medium**: Callbacks held in a `useRef`; the global keydown listener attaches once for the editor's lifetime — only `pagedEditorRef` (stable) remains in the dep array.

CC on behalf of @jan-kubica

PR #368, opened 2026-05-16
fix(case-law): narrow ingestion to dedicated stella_ingestion role

jan-kubica:
## Summary
- Case-law ingestion daemon ran through `createScopedDb` → `SET LOCAL ROLE stella`. The 2026-05-10 RLS bootstrap (`apps/api/drizzle/20260510140000_document_rls_role_bootstrap`) left `case_law_*` read-only for `stella` so customer-scoped paths cannot mutate the global corpus. Every adapter write failed with `permission denied for table case_law_decisions`; after `SUSTAINED_FAILURE_THRESHOLD` (5) consecutive failures each adapter halted and the daemon went silent. Observed today on prod: ~9 hours of zero ingestion across all adapters.

## Why a third role, not just granting writes to `stella`
- Granting writes to `stella` re-opens the surface the May 10 migration closed — any tenant code path that issues an UPDATE on `case_law_*` would succeed
- Going through the master connection expands the daemon's write surface to every table in the DB with RLS bypassed
- A narrowly-scoped role (writes only on `case_law_*`) preserves the loud `permission denied` failure mode if a future ingestion change accidentally touches a non-case-law table

## What changed
- **New migration `20260516000000_case_law_ingestion_role`** creates `stella_ingestion` NOLOGIN, grants SELECT on the nine `case_law_*` tables, grants INSERT/UPDATE/DELETE on the eight mutable ones (sources stays config-only), grants sequence usage, adds permissive RLS policies, and grants the migration connection role membership so `SET ROLE` works
- **New `createIngestionDb`** in `apps/api/src/db/scoped.ts` mirrors `createScopedDb` but switches to `stella_ingestion` and skips the `app.*` session settings (corpus has no tenant)
- **Daemon swap**: `ingest-case-law.ts` uses `createIngestionDb` for the per-cycle pipeline, the search-index backfill loop, and the ingestion-events insert. `ensureSource` keeps the master connection because `case_law_sources` writes are bootstrap-only and intentionally outside the daemon role

## Role matrix (after this PR)
| Role | Writes case_law? | Writes tenant tables? | RLS bypass? |
|---|---|---|---|
| `stella` (customer paths) | ❌ | ✅ with RLS scope | ❌ |
| `stella_ingestion` (daemon) | ✅ (8 tables) | ❌ | ❌ |
| master (migrations, source bootstrap) | ✅ | ✅ | ✅ |

## Test plan
- [x] `tsgo --noEmit` clean for `apps/api`
- [ ] db-migrations workflow validates the new migration applies clean against postgres:18 + replays journals
- [ ] After deploy: `case_law.ingestion.decision_failed` with `permission denied` stops appearing; `[<adapter>] Inserted: N` lines resume

## Deploy ordering
Drizzle migrate runs before the new task definition rolls (per `release.yml` + the migrate ECS task), so the role exists by the time the daemon container restarts. No special promotion needed.

### DIFF
--- a/apps/api/drizzle/20260516000000_case_law_ingestion_role/migration.sql
+++ b/apps/api/drizzle/20260516000000_case_law_ingestion_role/migration.sql
@@ -1,0 +1,127 @@
+-- Bootstrap `stella_ingestion`, the NOLOGIN role used by the case-law
+-- ingestion daemon to mutate the global corpus.
+--
+-- The 2026-05-10 RLS bootstrap left case_law_* tables read-only for
+-- the customer-scoped `stella` role so tenant code paths cannot
+-- mutate global reference data. Ingestion is an admin operation and
+-- needs write access — but going through the master connection
+-- would also grant writes on every tenant table (RLS bypassed,
+-- every grant inherited), so we add a third role narrowly scoped
+-- to the case-law corpus.
+
+CREATE ROLE stella_ingestion NOLOGIN;
+
+-- Schema access; mirrors the stella bootstrap.
+GRANT USAGE ON SCHEMA public TO stella_ingestion;
+
+-- SELECT on the nine case_law_* tables. Reads are needed for
+-- upsert logic, source cursors, and dedup checks. Deliberately
+-- nothing else — `permission denied for table foo` is the loud
+-- failure mode we want if a future ingestion change accidentally
+-- touches anything outside the corpus.
+GRANT SELECT ON TABLE
+  "case_law_sources",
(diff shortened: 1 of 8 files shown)


jan-kubica:
Both Codex findings addressed in 8ebded3:

- **P1 cursor update**: column-level `GRANT UPDATE (sync_cursor, last_sync_at, updated_at) ON case_law_sources TO stella_ingestion`. Drizzle's `.$onUpdate` on `updated_at` is also covered. Other source columns stay migration-managed.

PR #471, opened 2026-05-27
feat(chat): add searchInEntityContent tool

jan-kubica:
## Summary

- New `read.searchInEntityContent({ matterRefs, entityRefs, query, limit?, caseSensitive?, wholeWord? })` tool. Scans the FULL extracted plaintext of an entity and returns each match's character offset plus a ~200-char context window. `totalHits` is the absolute count; `hits` is capped at `limit` (default 10, max 20). Server-side scan is capped at 100 matches per entity to fence pathological queries.
- Prompt steers the model toward this tool for "where/whether/how is X mentioned" questions instead of `getMatterEntityContents`, which only returns the first 8 000 chars and silently truncates the rest.
- Defensive prompt note: `hits[].position` is a CHARACTER OFFSET, not a block id; do NOT fabricate `#folio:b-<position>` chips from search positions. (Stop-gap until block-id-aware search lands — see follow-up.)

## Why

Long-document lookups silently fail today. Asking "how is X defined?" against a 50-page contract returns the cover page + Index of Defined Terms (the first 8 000 chars), and the model honestly reports "it isn't defined here" — because the definition section was discarded by the server before the model ever saw it.

## Follow-up

This PR returns char offsets only. The proper end-state is to also return Word `w14:paraId` per hit so the model has a canonical anchor for `#folio:<paraId>` chips. That requires:
- Porting eigenpal's `ParaIdAllocatorExtension` so every paragraph has a paraId.
- Switching `createFolioAIEditSnapshot` to key anchors by paraId.
- Recording per-paragraph offsets during extraction (new `extracted_blocks` JSONB column).

Tracked as a separate PR.

## Test plan

- [x] `bun --filter @stll/api typecheck` / `lint` clean.
- [x] Existing manifest tests updated to assert the new contract is registered.
- [ ] Manual: ask "how is Blue Sky Laws defined?" against a long contract — model returns the actual definition snippet instead of "only in the index".


jan-kubica:
Addressed the chatgpt-codex review (P2 on `workspace-function-registry.ts:93` — scan-cap honesty) in `e5ed6a986`, then took the cleaner cut in `e0a489732` per follow-up feedback:

1. **Removed `hits[].position` entirely.** Char offsets were dead weight: the only place they'd plausibly be used is a chip href (`#folio:b-<position>`), and that's a fabricated id that doesn't resolve. Hits are now ordered by document position — array index is enough for relative ordering. Drops the matching prompt warning too.
2. **`totalHitsCapped: boolean`** still flags when the defensive scan cap was reached, so the model treats `totalHits` as a lower bound rather than misreporting "exactly 100 matches" when the count is higher. Schema descriptions spell out both fields.

End-state: `searchInEntityContent` returns `{ snippet }[]` + `totalHits` + `totalHitsCapped` + `truncated` (hits-array-clipped) per entity. The cleaner follow-up — having the server return a real `paraId` per hit so chips Just Work — depends on PR #473 (canonical paraId) + the extraction-pipeline-records-offsets follow-up. Tracked in #473's body.

CC on behalf of @jan-kubica

PR #473, opened 2026-05-27
feat(folio): use Word w14:paraId as canonical AI block anchor

jan-kubica:
## Summary

- Port eigenpal upstream's `ParaIdAllocatorExtension` (Apache-2.0, unchanged behaviour) and the `findParagraphByParaId` / `findStartPosForParaId` helpers. Add `hexId` generator with the same OOXML upper-bound constant.
- Register the allocator in `StarterKit` ahead of the change tracker so any freshly-allocated id is already on the paragraph when downstream effects observe it. `addToHistory: false` keeps undo/redo clean.
- Switch `createFolioAIEditSnapshot` to anchor blocks by Word `w14:paraId` (`b-${paraId}`) when the attribute is present. Falls back to the existing sequential ordinal for the pre-allocator case (snapshots of states without the plugin, non-paragraph textblocks) so existing callers keep working.

## Why

Sequential block ids (`b-0001`, `b-0002`, …) are not stable across structural edits. Inserting a paragraph at the top of the document renumbers every block after it, so any AI-issued blockId from a prior turn silently points at the wrong paragraph in the next. The pathology surfaces in chat as "the chip doesn't work" — `scrollToBlock` succeeds against a real-looking id that no longer maps to the intended block.

Word already gives us a canonical per-paragraph anchor in OOXML — `w14:paraId`, an 8-char hex value assigned at paragraph creation. Folio already parsed it (`paragraphParser.ts:1684`) and exposed it on the paragraph node (`ParagraphExtension.ts:324`), but the AI snapshot ignored it.

## Where this came from

Eigenpal hit the same wall in their agent toolkit and migrated to a "locate, then mutate" architecture anchored on `paraId`. Their CHANGELOG: *"paraIds survive concurrent edits and tool-loop iterations; ordinal indices don't."* This PR lifts the three pieces verbatim (NOTICE-tagged in each file's header):

- `packages/core/src/utils/hexId.ts`
- `packages/core/src/prosemirror/utils/findParagraphByParaId.ts` / `findStartPosForParaId.ts`
- `packages/core/src/prosemirror/extensions/features/ParaIdAllocatorExtension.ts`

## Follow-up (separate PR)

This makes EDIT-mode chips stable across turns. The READ-mode case (chips from `searchInEntityContent` results) still needs the extraction pipeline to record per-paragraph offsets so the server can return a real `paraId` alongside each char-offset hit. Tracked separately.

## Test plan

- [x] `bun --filter @stll/folio test` — 915 pass, 0 fail (+11 new tests covering hex bound / allocator behaviour / helper lookups).
- [x] `bun --filter @stll/folio typecheck` clean.
- [x] `bun --filter @stll/folio lint` clean.
- [x] `bun run typecheck` across all 23 packages clean.
- [ ] Manual: chip targeting a paragraph that has shifted position after a top-of-doc insertion still scrolls to the correct paragraph.


jan-kubica:
Addressed the chatgpt-codex review (P2 on `snapshot.ts:65` — apply path resolved by hash+ordinal instead of the paraId in the new block id) in `b0fc7672f`, plus a cleaner unification per follow-up feedback:

1. **Dropped the `b-` prefix.** Snapshot now emits the bare `w14:paraId` (8-char uppercase hex) as the canonical block id everywhere — `apply-active-docx-edits` blockIds, `#folio:<paraId>` chip hrefs, `scrollToBlock`. Pre-allocator fallback (paragraphs without a `paraId`) is now `seq-${ordinal}` so the apply path can tell at a glance which lookup strategy to use.
2. **Wired the apply path through a paraId-direct index.** `collectLiveBlocksByParaId` builds an O(1) lookup; `resolveOperation` prefers it when the blockId is 8-char hex. The hash+ordinal path is now the FALLBACK, only used for `seq-` ids and for stale paraId refs that no longer match a live block.
3. **Regression test added** (`paraId-anchored op resolves the right block when a same-text duplicate appears before it`) — covers exactly the scenario from your review: a `Payment.` duplicate inserted before the original target between snapshot and apply. With the old hash+ordinal-only path this would silently mutate the wrong block; with the paraId-direct path the duplicate stays untouched.

Prompt updates: example blockIds switch from `b-0010` to `1A2B3C4D` style; chip example updated to `#folio:1A2B3C4D`.

CC on behalf of @jan-kubica

PR #475, opened 2026-05-27
feat(workspaces): retry one AI-extracted cell via right-click

### LINKED ISSUE #307: feat: retry table result
Allow user to manually retry an individual result in the table (ie the value in the cell), optionally (via dropdown?) with different model. 

Maybe right click on the cell > Menu is the right place to put this, as well as the Metadata column in the file viewer, which also shows the extracted data.

jan-kubica:
## Summary
- New `POST /workspaces/:workspaceId/cell-retry`: validates the target property is AI-extracted and the cell isn't locked, resets the field to `pending`, then enqueues a workflow scoped to that single property.
- `startWorkflow` gains an optional `propertyIds` filter so the execution plan can be narrowed to one cell without touching the rest of the entity.
- Right-click on an AI-model table cell now exposes "Retry" via the existing `useAnchoredMenu` primitive (reuses `common.retry`, no new i18n key).

Concurrent full-workspace workflows still take the per-workspace Redis lock, so a retry started during an active run returns `already-running`; the user retries once the run finishes.

Closes #307.

### DIFF
--- a/apps/api/src/handlers/workspaces/cell-retry.ts
+++ b/apps/api/src/handlers/workspaces/cell-retry.ts
@@ -2,13 +2,12 @@ import { Result } from "better-result";
 import { and, eq } from "drizzle-orm";
 import { t } from "elysia";
 
-import { cellMetadata, fields } from "@/api/db/schema";
+import { cellMetadata } from "@/api/db/schema";
 import { createSafeHandler } from "@/api/lib/api-handlers";
 import type { HandlerConfig } from "@/api/lib/api-handlers";
-import { createSafeId } from "@/api/lib/branded-types";
 import { tSafeId } from "@/api/lib/custom-schema";
 import { HandlerError } from "@/api/lib/errors/tagged-errors";
-import { startWorkflow } from "@/api/lib/workflow-queue";
+import { isWorkflowRunning, startWorkflow } from "@/api/lib/workflow-queue";
 
 const config = {
   permissions: { workspace: ["update"] },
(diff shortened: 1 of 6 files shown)


jan-kubica:
Addressed bot review in 94f01d002:

- The stuck-pending race is now structurally impossible: `prepareBatch` takes a `forcedPropertyIds` set, so a workflow scoped via `startWorkflow.propertyIds` re-processes the requested cells without the handler having to pre-write `pending`. Dropped the snapshot/restore tangle that approach would have needed.
- Handler now does an upfront `isWorkflowRunning` check and returns 409, and translates `startWorkflow`'s `already-running` / `failed` results into 409 / 500 instead of a misleading 200.
- Right-click menu disables the retry item when the cell is locked or the entity is read-only, and blocks double-submits while the request is in flight.

CC on behalf of @jan-kubica.

jan-kubica:
Addressed Codex P1 in 8e5142e6a.

`startWorkflow` now sets a `scoped` flag in Redis when called with `propertyIds`. `finishWorkflow` checks it before the freshen step and the straggler follow-up — both are skipped on scoped runs, so the property's workspace-wide stale marker is preserved and the next full extraction will still re-process every other entity's cell for that property. TTL is refreshed alongside the other lock keys so a long-running retry can't drop the flag mid-flight.

CC on behalf of @jan-kubica.

jan-kubica:
Addressed Codex P2 in 99b2e7c64.

`useRetryCell` now routes server rejections through `userErrorMessage` into `stellaToast.error`, so locked cells / read-only entities / concurrent-workflow 409s surface as a toast instead of silently closing the menu. 5xx still falls back to `errors.actionFailed`.

CC on behalf of @jan-kubica.

PR #483, opened 2026-05-28
fix(docx): break infinite render loop in inspector file open

jan-kubica:
## Summary

- Opening any file in the inspector or peek view crashes with React error #185 (\"Maximum update depth exceeded\"). Regressed in #476 by the change to `useDocxFitZoom` that dropped the dependency array on its `useLayoutEffect` and guarded re-attachment with a `trackedRef`.
- The cleanup nulls `trackedRef.current` before the next effect runs, so the guard `container === trackedRef.current` evaluates false on every commit. The observer re-attaches and `updateZoom()` runs synchronously each render. With `editor.setZoom(targetZoom)` toggling the vertical scrollbar (changing container `clientWidth` by ~15px), `setFitZoom` oscillates until React aborts.
- Swap the `RefObject` + `useLayoutEffect` shape for a callback ref returned from the hook. React invokes it exactly once when the node attaches and runs the returned cleanup on detach, so the ResizeObserver is created per node lifetime regardless of parent re-renders. The original "container appears late after the loading shell" issue (which the #476 rewrite was trying to solve) is handled natively by the callback-ref form.
- Callers that need the DOM node elsewhere (wheel zoom, child queries) keep their `useRef` and compose it with the hook's callback ref via `composeRefs` from `lib/slot` (now exported).

## Test plan

- [ ] Upload a fresh DOCX in a workspace, open it from Documents — inspector renders the file without throwing.
- [ ] Open the same file in peek mode (mention from chat / direct route) — no console errors.
- [ ] DOCX inspector tab on a narrow pane (< 600px) — fit zoom settles, no flicker.
- [ ] Resize the inspector pane — zoom recomputes smoothly.
- [ ] Ctrl/Cmd + wheel inside the docx — wheel zoom still works (unchanged `useDocxWheelZoom`).
- [ ] PDF peek (non-docx path) — unchanged.

## Follow-up

A Playwright smoke covering \"upload file → open in inspector → assert console clean\" would have caught this. None exists today; worth adding in a separate PR.

CC on behalf of @jan-kubica

### DIFF
--- a/apps/web/src/lib/slot.tsx
+++ b/apps/web/src/lib/slot.tsx
@@ -39,7 +39,7 @@ type SlotProps = React.HTMLAttributes<HTMLElement> & {
  * callbacks and returns a combined cleanup so React can
  * invoke it on unmount instead of re-calling with `null`.
  */
-const composeRefs =
+export const composeRefs =
   <T,>(
     ...refs: (React.Ref<T> | undefined)[]
   ): ((node: T | null) => (() => void) | undefined) =>
--- a/apps/web/src/routes/_protected.workspaces/$workspaceId/$viewId.document.tsx
+++ b/apps/web/src/routes/_protected.workspaces/$workspaceId/$viewId.document.tsx
@@ -3,6 +3,7 @@ import {
   Suspense,
   useEffect,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
 } from "react";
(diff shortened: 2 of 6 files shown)


jan-kubica:
Both gemini and codex flagged the same issue: the inline `composeRefs(containerRef, fitZoomRef)` callback ref was unstable across renders, reintroducing the same #185 loop the PR is meant to fix.

Confirmed live and addressed in 6eb7d6a8 — the three call sites (`peek-pdf-viewer.tsx`, `docx-browser-editor.tsx`, `$viewId.document.tsx`) now wrap the composition in `useMemo`, so React sees a stable callback ref and attaches the underlying `ResizeObserver` exactly once per node lifetime.

CC on behalf of @jan-kubica

PR #514, opened 2026-05-30
fix(folio): anchor inline images at line top to prevent clipping

jan-kubica:
## Summary

Ports the inline-image clipping sub-fix of upstream eigenpal/docx-editor #409 to folio. The layout painter's `renderInlineImageRun` set `img.style.verticalAlign = "middle"`, which sits an inline image half above / half below the text baseline; the top half visually escapes the paragraph's bounding box and clips against borders or the next line. Setting `verticalAlign = "top"` anchors the image at line top regardless of CSS reset (notably Tailwind preflight's `img { vertical-align: middle }` global rule).

Scope is intentionally limited to the verticalAlign literal change. The other two #409 sub-items (top-of-page `spaceBefore`, break-only page break) are already handled in folio and are not touched here.

## Test plan

- [x] Added regression test in `packages/folio/src/core/layout-painter/renderParagraph-inline-image.test.ts` — `anchors inline images at line top` — asserting an `` rendered through `renderLine` ends up with `style.verticalAlign === "top"`. Confirmed RED before the fix, GREEN after.
- [x] `bun --filter @stll/folio test` — 1003 pass / 0 fail (1002 before, +1 regression test).
- [x] `bun --filter @stll/folio lint` — clean.
- [x] `bun --filter @stll/folio typecheck` — clean.

### DIFF
--- a/packages/folio/src/core/layout-painter/renderParagraph-inline-image.test.ts
+++ b/packages/folio/src/core/layout-painter/renderParagraph-inline-image.test.ts
@@ -107,6 +107,45 @@ describe("renderLine inline image handling", () => {
     expect(imageEl?.style.height).toBe("29px");
   });
 
+  // Regression for eigenpal #409 inline-image clipping sub-fix. Tailwind
+  // preflight sets `img { vertical-align: middle }` globally; the JS literal
+  // must explicitly anchor inline images at line top so they don't sit
+  // half above / half below the baseline (which clips against paragraph
+  // borders or following lines).
+  test("anchors inline images at line top", () => {
+    const imageRun: ImageRun = {
+      kind: "image",
+      src: "data:image/png;base64,",
+      width: 120,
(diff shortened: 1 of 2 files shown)


jan-kubica:
CC on behalf of @jan-kubica

Closing — this change is a regression rather than a fix. Re-checking upstream PR #409 in detail:

- The PR's description mentions `verticalAlign='top'` as an intent but the actual code change ends with `img.style.verticalAlign = 'middle'` plus an explanatory comment noting that Tailwind preflight already sets this globally and that the inline-block + flex-centering branch handles image-only lines.
- Upstream `main` currently uses `'middle'` in the inline-image render path; folio already matches.

The real #409 inline-image clipping fix is the flex-centering branch for image-only lines, which folio may or may not need separately. That requires a more targeted probe rather than the verticalAlign change here.

PR #561, opened 2026-05-30
feat(business-registries): add Slovak RPO v1 adapter

jan-kubica:
## Summary

Adds `@stll/business-registries/rpo` for Slovak entity lookups via the Štatistický úrad SR `api.statistics.sk/rpo/v1/` public REST surface (CC BY 4.0, no auth) and wires it into the unified `business_registry_lookup` chat tool through the shared dispatch module. SK IČO (8 digits, MOD-11) reuses `@stll/stdnum/sk/ico`, which re-exports the Czech IČO validator unchanged — both schemes inherit from the pre-1993 Czechoslovak register.

Two upstream-shape notes worth flagging for review: (a) the documented v2 endpoint is currently 404, so the adapter targets v1; (b) `/entity/{id}` requires the upstream's internal numeric `id`, not the IČO, so canonical-ID lookup is a two-step fetch (search by `identifier` → detail by internal id) — the search row alone omits `legalForms`, `activities`, `statutoryBodies`, and `statisticalCodes`.

Prior-art investigation: only one TS client surfaced (`ShepDogSk/rpo-mcp-server`, an MCP wrapper); no maintained typed SDK on npm. Wrote fresh against the documented v1 REST surface.

## Test plan
- [x] `bun --filter @stll/business-registries test src/rpo` (28 pass)
- [x] `SMOKE_TEST=1 bun --filter @stll/business-registries test src/rpo/client.test.ts` (12 pass against live API)
- [x] `bun run lint` (clean)
- [x] `bun run typecheck` (clean)
- [x] `bun run i18n:check` (no new keys needed; chat tool is unified)
- [x] `bun run test` full suite (2085 API tests pass)


jan-kubica:
Draft per maintainer request: holding for v1-vs-v2 endpoint investigation. PR currently targets `api.statistics.sk/rpo/v1/` (the only working public path); v2 path returns 404. Will revisit once we have a confirmed v2 URL (private/preview or otherwise). The 2-step lookup (search by IČO → fetch by internal ID) is approved.

jan-kubica:
Superseded by the ORSR adapter — see the replacement PR (about to open) targeting `sluzby.orsr.sk` instead of the Statistical Office's RPO meta-register. ORSR exposes statutory bodies, share capital, court files, and the trade-register file reference; RPO v1 does not.

PR #679, opened 2026-06-08
feat(folio): preserve run shading (w:shd) backgrounds via a strict mark

jan-kubica:
## Problem

Folio models `w:highlight` as a strict OOXML named-palette union (no hex) — `shading` (`w:shd`) is a separate field. Arbitrary run-background fills (a `w:shd` from Word/Google Docs, or a custom highlight another editor saved as shading) were parsed into `formatting.shading` but **dropped at ProseMirror conversion**: no mark carried them, so the background silently vanished on reload, and the run serializer never re-emitted it.

Upstream eigenpal #722 (#712) maps such shading onto the highlight mark as hex. That does not fit folio's strict named-palette highlight (the naive port produces an invalid document — validation rejects hex). So this is folio's idiom for the same fix.

## Change

A dedicated `runShading` ProseMirror mark carries the shading **fill** as a flattened `ColorValue` (rgb / theme*) plus a non-default `pattern`, mirroring the `textColor` mark. The shading now round-trips (import → PM → export) and renders.

- **Schema/validation:** new `RunShadingExtension` (no `parseDOM` — the highlight mark already claims `background-color` on paste; no toolbar command — this is import-only, like the `rtl` mark). `readRunShadingMarkAttrs` validates rgb/theme/pattern; unlike `highlight` it accepts arbitrary hex.
- **Conversion:** `toProseDoc` creates the mark from `formatting.shading`; `fromProseDoc` rebuilds `formatting.shading` so the run serializer re-emits `w:shd`. Shared converters in `runShadingMark.ts`.
- **Render:** `toFlowBlocks` resolves the fill to a CSS background (`RunFormatting.shading`); the painter paints `run.highlight ?? run.shading` (an explicit highlight wins).
- The default `clear` pattern is never stored — the fill renders solid and re-serializes to `w:val="clear"`.

## Separate fix (own concern): `resolveShadingFill` clear handling

`resolveShadingFill` returned transparent for any `w:val="clear"` shading **before** inspecting the fill. Per ECMA-376 §17.18.78 only `nil` means "no shading"; `clear` means "no pattern", so the `w:fill` colour shows as a solid background. The early return dropped legitimate `<w:shd w:val="clear" w:fill="…"/>` backgrounds. This also affects paragraph/table run shading that flows through `textToStyle`. White/`auto` fills stay transparent (page-background no-ops); paragraph **block** shading uses its own path and is untouched.

## Tests

- `issue-712-run-shading-roundtrip.test.ts`: import → render → export round-trip, gating (auto / theme-only / fill-less / clear-vs-non-clear pattern), highlight-wins precedence at the flow layer.
- `renderParagraph-run-shading.test.ts`: painter paints shading as a background; highlight wins on the same run.
- `formatToStyle-shading.test.ts`: the `clear`/`nil` fix across all fill combinations.

## Notes

Touches `renderParagraph.ts` (the run-background block only) — disjoint from the RTL/list-marker painter PRs in this sync, so it rebases cleanly in any merge order. Runtime toolbar/selection mark switches are intentionally not extended (safe `default` branches; the mark is import-only and survives via to/fromProseDoc).

### DIFF
--- a/packages/folio/src/core/layout-bridge/toFlowBlocks.ts
+++ b/packages/folio/src/core/layout-bridge/toFlowBlocks.ts
@@ -51,6 +51,7 @@ import {
   expectMathAttrs,
   expectParagraphAttrs,
   expectRunFormattingOverrideMarkAttrs,
+  expectRunShadingMarkAttrs,
   expectTableAttrs,
   expectTableCellAttrs,
   expectTableRowAttrs,
@@ -60,6 +61,7 @@ import {
   expectTrackedChangeMarkAttrs,
   expectUnderlineMarkAttrs,
 } from "../prosemirror/attrs";
+import { runShadingAttrsToShading } from "../prosemirror/conversion/runShadingMark";
 import type { RunFormattingOverrideAttrs } from "../prosemirror/schema/marks";
 import type {
   ImageAttrs,
@@ -75,6 +77,7 @@ import type {
 } from "../types/document";
 import { NUMBER_FORMAT_VALUES } from "../types/documentEnumValues";
 import { resolveColor, resolveHighlightToCss } from "../utils/colorResolver";
+import { resolveShadingFill } from "../utils/formatToStyle";
 import {
   pointsToPixels,
   halfPointsToPixels,
@@ -470,6 +473,17 @@ function extractRunFormatting(
           expectHighlightMarkAttrs(mark).color,
         );
         break;
+
+      case "runShading": {
+        const shadingCss = resolveShadingFill(
+          runShadingAttrsToShading(expectRunShadingMarkAttrs(mark)),
+          theme,
+        );
+        if (shadingCss) {
+          formatting.shading = shadingCss;
+        }
+        break;
+      }
 
       case "fontSize": {
         const attrs = expectFontSizeMarkAttrs(mark);
--- a/packages/folio/src/core/layout-engine/types.ts
+++ b/packages/folio/src/core/layout-engine/types.ts
@@ -41,6 +41,12 @@ export type RunFormatting = {
   // `#FFFF00` are not union members. Either keep a separate CSS field or move
   // resolution to the painter before this can become the named-color union.
   highlight?: string;
+  /**
+   * Run-level shading (w:shd) used as a background, stored as a *resolved CSS*
+   * color (like `highlight`). Painted only when there is no `highlight` (an
+   * explicit highlight wins). eigenpal #722 (#712).
+   */
+  shading?: string;
   fontFamily?: string;
   fontSize?: number;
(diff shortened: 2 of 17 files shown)


jan-kubica:
Addressed the review feedback:

- **toDOM theme-vs-rgb (gemini):** confirmed `resolveColor` checks `themeColor` before `rgb`, so a fill carrying both would resolve the theme slot — and with `theme=null` in the overlay that lands on the Office default (often black). `toDOM` now uses the concrete `rgb` directly; theme-only fills (no rgb) are left to the paged painter, which has the real theme.
- **solid pattern without a fill (codex):** `<w:shd w:val="solid" w:color="…"/>` now flattens the pattern color into the fill so the background renders and round-trips, instead of being dropped by the fill guard. Test added.
- Lint: removed an unnecessary optional chain and suppressed the hardcoded-color rule in the colour-assertion test fixtures.

jan-kubica:
Addressed: added a `background-color` parseDOM to the runShading mark. It defers to the highlight mark for any color in the OOXML named palette and claims the rest (arbitrary hex/rgb) as run shading, so pasting a custom background no longer drops it. (parseDOMHighlightColor/normalizeCssColorKey are now exported from HighlightExtension to keep the two marks from double-claiming a color.)

jan-kubica:
Addressed: for a `solid` pattern the pattern color (`w:color`) is painted over the fill, so it's the visible background. `shadingToRunShadingAttrs` now prefers `color ?? fill` for solid patterns (instead of always choosing the fill), so `<w:shd w:val="solid" w:color="FF0000" w:fill="00FF00"/>` renders FF0000. Test added.

jan-kubica:
Addressed: a non-clear pattern's foreground color (`w:color`, e.g. `pct25 w:color=FF0000`) is now carried as a `patternColor` attr and restored to `shading.color` on read-back, so it round-trips on export instead of being dropped. Test added.

jan-kubica:
Addressed: in ProseMirror's DOM serialization the later-registered mark is the inner span (its background wins). Registered `runShading` *before* `highlight` in the StarterKit so an explicit highlight wins the editor-overlay background when a run carries both — matching the paged painter's `run.highlight ?? run.shading`.

jan-kubica:
@codex review

jan-kubica:
Merging on confidence. All review feedback has been addressed across Codex's six review rounds (the latest being the runShading-before-highlight DOM-precedence reorder in ce4e26707), CI is fully green, and the change is well-tested. Codex hasn't re-reviewed the current head after the rebase force-push (the `@codex review` nudge didn't re-trigger it in ~40 min), and the only delta since its last review is exactly the reorder it requested, so there's nothing outstanding to resolve.

PR #681, opened 2026-06-08
fix(folio): hang list markers into the margin when hanging > left indent

jan-kubica:
## Problem

A numbered list whose direct paragraph indent has `hanging` greater than `left` must hang its marker into the left margin (as Word does). Folio clamped the marker position with `Math.max(0, indentLeft - hanging)`, pinning it to the content edge so the numbers shifted right of the text above.

## Change

Computes a signed `markerStart`:

- `hanging > 0` → `indentLeft - hanging` (can be negative)
- else → `indentLeft + firstLine`

`padding-left` clamps to `≥ 0` (CSS padding can't be negative); when `markerStart < 0` the negative portion rides on the marker element's own `margin-left`. `text-indent` stays `0` (Chrome folds a negative text-indent into the marker's inline-block box and breaks its min-width slot).

Gated to `indentLeft > 0`: with no left indent the body/continuation lines already sit at `hanging`, so hanging the first-line marker into the margin there would misalign it. The `hanging <= left` and first-line-indent paths are unchanged.

## Test

`renderParagraph-list-hanging-indent.test.ts` — `hanging > left` hangs via negative margin-left (exact px), `hanging <= left` unchanged (padding, no margin), `left == 0` gated off.

## Notes

Touches `renderParagraph.ts` (the first-line list-marker block only) — disjoint from the RTL and run-shading painter PRs in this sync, so it rebases cleanly in any merge order.

### DIFF
--- a/packages/folio/src/core/layout-painter/renderParagraph-list-hanging-indent.test.ts
+++ b/packages/folio/src/core/layout-painter/renderParagraph-list-hanging-indent.test.ts
@@ -1,0 +1,181 @@
+// eigenpal #730 (#729) — a numbered list whose direct paragraph indent has
+// `hanging` greater than `left` must hang its marker into the left margin (as
+// Word does), not clamp it to the content edge. The marker line keeps
+// `text-indent: 0` and the hang comes from padding-left; CSS padding can't be
+// negative, so when `left - hanging < 0` the negative portion rides on the
+// marker's own `margin-left`. Without it the old `Math.max(0, left - hanging)`
+// clamp pinned the marker to the content edge, shifting the numbers right of
+// the text above.
+
+import { describe, expect, test } from "bun:test";
+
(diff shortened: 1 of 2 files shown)


jan-kubica:
Thanks — the RTL list-marker mirroring (apply the offset on `padding-right`/`margin-right` and gate on the start-side indent for bidi paragraphs) is a valid improvement, but it's a pre-existing concern that's out of scope for this PR: the marker was positioned LTR-only before this change too (the old `Math.max(0, indentLeft - hanging)` also wrote `paddingLeft`). This PR only fixes the clamp so a marker can hang into the margin; it doesn't change RTL handling.

I'd rather land the RTL marker mirroring alongside the RTL base-direction work (#680) so the start/end-side logic is added once, consistently, rather than partially here. Filed mentally as a follow-up; happy to do it there.

PR #722, opened 2026-06-12
fix(web): source saved prompts org from the auth context

jan-kubica:
`useSavedPrompts` read `activeOrganizationId` via `getRouteApi("/_protected").useRouteContext`, which throws `Could not find an active match` when the hook renders outside that route tree — concretely, the public law decision workspace mounts the inspector chat panel for authenticated users (whenever persisted inspector tabs exist), taking the whole page to the error boundary. Source the organization from `useAuthenticatedUser()` instead, which both the `/_protected` tree and the public law workspace provide; the panel itself already uses it.

### DIFF
--- a/apps/web/src/lib/authenticated-user-context.tsx
+++ b/apps/web/src/lib/authenticated-user-context.tsx
@@ -34,3 +34,12 @@ export const useAuthenticatedUser = (): AuthenticatedUser => {
   }
   return user;
 };
+
+/**
+ * Like useAuthenticatedUser, but returns null outside the provider.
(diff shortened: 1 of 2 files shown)


jan-kubica:
Updated: `useSavedPrompts` now uses a new null-tolerant `useMaybeAuthenticatedUser()` and returns an empty list (query disabled) when no user context exists, so AI surfaces shown to anonymous visitors before signup can mount it without crashing; account-bound prompts simply do not appear. The strict `useAuthenticatedUser` stays for genuinely authenticated-only code.

CC on behalf of @jan-kubica

PR #742, opened 2026-06-13
chore(business-registries): single-source ID checksums to @stll/stdnum

jan-kubica:
Replace the hand-rolled check-digit algorithms in four registry adapters with the corresponding validators from `@stll/stdnum`, matching the existing `ares`/`orsr` pattern:

- `brreg` — NO organisasjonsnummer (`no/orgnr`)
- `prh` — FI y-tunnus (`fi/ytunnus`)
- `recherche-entreprises` — FR SIREN/SIRET (`fr/siren`, `fr/siret`)
- `gcis` — TW 統一編號 / UBN (`tw/ubn`)

Each adapter keeps its own normalizer, which owns the API/canonical form: `prh` retains the hyphenated-format guard (stdnum also accepts the bare 8-digit form), and the FR/`gcis` normalizers keep newline handling that stdnum's `compact` does not.

The `edgar`, `companies-house`, and `krs` adapters have no check digit and `vies` is format-only, so every checksum-bearing adapter now sources its algorithm from `@stll/stdnum`.

Behaviour change: the FR La Poste head office (`35600000000048`) now validates via standard Luhn instead of being rejected by the digit-sum carve-out.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -365,7 +365,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@stll/country-codes": "workspace:*",
-        "@stll/stdnum": "^2.0.1",
+        "@stll/stdnum": "^2.1.1",
       },
       "devDependencies": {
         "@stll/typescript-config": "workspace:*",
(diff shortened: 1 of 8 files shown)


jan-kubica:
Bot review addressed:

- **Codex (dotted FR ids):** real bug, fixed. `normalizeSiren` now reuses `@stll/stdnum`'s `compact`, so a valid spaced/dotted SIREN/SIRET canonicalizes to the digits used for the API query and the exact-match check instead of resolving to `null`. New `normalizeSiren` test covers it.
- **Gemini (normalize before validate):** the underlying gap was a missing whitespace strip in `@stll/stdnum`'s `compact`, not a consumer concern. Fixed upstream and released as `@stll/stdnum@2.1.1` (now the required dependency), so the validators stay as thin delegations with no local wrapper.

PR #753, opened 2026-06-18
fix(folio): pagination fidelity for spaced paragraphs and section-break orientation

jan-kubica:
Ports two upstream `eigenpal/docx-editor` layout-engine correctness fixes into folio. Independent of the sibling sync PRs (disjoint files).

### Changes
- **Spaced-paragraph pagination** (`eigenpal/docx-editor#782`): a multi-line paragraph after a block with a large `space-after` wrongly jumped a full page. The line-fit loop reserved only the paragraph's own `spaceBefore`, while `addFragment` reserves the spacing collapsed with the previous block's trailing spacing (`max(spaceBefore, trailingSpacing)`). The loop now reserves the same collapsed amount, so the first fragment splits at the page boundary instead of being punted whole.
- **Orientation-changing continuous section break** (`eigenpal/docx-editor#841`): a `continuous` break that changes page size/orientation is promoted to a page break, since the new geometry cannot share a physical sheet with the preceding section (Word/LibreOffice behavior). Adapted to folio's flow-anchored section model.

### Tests
- `paragraphSplitWithSpaceBefore.test.ts`: first fragment reserves collapsed trailing spacing.
- `continuousSectionGeometry.test.ts`: orientation-changing continuous break promotes to a page break.

Full folio suite green.

### DIFF
--- a/packages/folio/src/core/layout-engine/continuousSectionGeometry.test.ts
+++ b/packages/folio/src/core/layout-engine/continuousSectionGeometry.test.ts
@@ -121,5 +121,132 @@ describe("continuous section break geometry", () => {
       page.fragments.some((f) => f.kind === "paragraph" && f.blockId === "a"),
     );
     expect(pageWithA?.size).toEqual({ w: 800, h: 1000 });
+    // The promoted break starts the next section, so the new page carries the
+    // next section's index (and thus its header/footer references).
+    expect(pageWithA?.sectionIndex).toBe(0);
+    expect(pageWithB?.sectionIndex).toBe(1);
+  });
+
+  test("a leading size-changing continuous break does not strand a blank page", () => {
+    // With no content laid out yet there is no sheet to share, so the break
+    // defers instead of materializing a blank page just to compare geometry;
+    // the first content opens directly on a new-geometry page.
+    const sectionBreak: SectionBreakBlock = {
+      kind: "sectionBreak",
+      id: "sb",
+      type: "continuous",
+      pageSize: { w: 800, h: 1000 },
+      margins: { top: 50, right: 50, bottom: 50, left: 50 },
+    };
+    const body = paragraph("b", 200);
+
+    const blocks: FlowBlock[] = [sectionBreak, body.block];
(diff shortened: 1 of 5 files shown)


jan-kubica:
@codex review

jan-kubica:
Round 2 addressed: when a paragraph cannot start on the current page *only* because its leading spacing collapses with the previous block's trailing spacing (spacing the next page sheds), the fit loop now advances first, so the paragraph re-fits whole on the next page. This avoids the `fittingLines === 0` fallback stranding its first line and `addFragment` carrying it over alone, which split a paragraph that fits whole into two same-page fragments (an artificial intra-page continuation). Guarded on `trailingSpacing > 0`, which an advance resets to 0, so it cannot loop. Regression test added.

jan-kubica:
@codex review

jan-kubica:
Round 3:

- **Infinite loop on oversized collapsed spacing (`index.ts:526`):** fixed, good catch. The advance-first branch now also requires `state.cursorY !== state.topMargin`. At a fresh page top `ensureFits` deliberately does not advance oversized content, so continuing there re-entered forever; the guard restricts the advance to pages that already hold content (where advancing genuinely sheds the trailing spacing), and otherwise falls through to the `fittingLines === 0` fallback. Added a regression test that terminates under a tight timeout.
- **Reuse blank pages before promoting section breaks (`index.ts:1328`):** this is a pre-existing, cross-cutting behaviour — the forced `nextPage`/`evenPage`/`oddPage` promotions (lines ~1270/1281/1297) run the identical `updatePageLayout -> startSection -> forcePageBreak({ coalesceBlankPage })` sequence; the continuous case here just mirrors it. A correct fix needs a paginator primitive that re-adopts the active layout onto an existing blank sheet (re-syncing page geometry, footnote reservation, and section page numbering) applied uniformly across all four break types. That belongs in a focused paginator change rather than this sync PR, so I've left it out here and noted it for follow-up.

jan-kubica:
@codex review

jan-kubica:
@codex review

CC on behalf of @jan-kubica

PR #754, opened 2026-06-18
fix(folio): painter rendering fidelity (superscript, justify, anchored images, HF overlay, empty-line selection)

jan-kubica:
Ports five upstream `eigenpal/docx-editor` painter fixes into folio. Independent of the sibling sync PRs (disjoint files).

### Changes
- **Superscript line height** (`eigenpal/docx-editor#846`): super/subscript now raise via a paint-only `position: relative` offset instead of `vertical-align`, which grew the CSS line box. A line carrying a footnote anchor no longer exceeds the measurer-reserved height. Folio keeps its run-derived raised font size (better than upstream's hardcoded ratio).
- **Justify + first-line indent** (`eigenpal/docx-editor#868`): the justify box is the full content width on every line; the first-line shift rides on `text-indent` only. Previously the box was also narrowed by `firstLine`, double-counting the indent so the first line fell short of the right margin.
- **Anchored-image alignment** (`eigenpal/docx-editor#787`, issue #777): an image alone on a line uses its own `wp:positionH` alignment (defaulting to left like Word), not the paragraph `jc`. `inside`/`outside` map to left/right.
- **Header/footer shape overlay** (`eigenpal/docx-editor#869`, #856): a page/margin-anchored shape (full-page letterhead) no longer inflates the header/footer interactive box over the body. The box tracks the in-flow band; header/footer content is `pointer-events: none` in normal mode and re-enabled while editing. Adapted to folio's flow-anchored footer.
- **Empty-paragraph selection sliver** (`eigenpal/docx-editor#836`, #776): drag-selecting across a blank line paints a fixed-width caret-height sliver. Fixed in folio's model-based `selectionToRects` (body + table-cell loops), where a zero-length line produced no rect.

### Tests
- `renderParagraph-line-box.test.ts`, `renderParagraph-justify-first-line-indent.test.ts`, `renderUtils-image-line-align.test.ts`, `renderPage.test.ts`, `selectionRects.test.ts`.

Full folio suite green.

### DIFF
--- a/packages/folio/src/core/layout-bridge/selectionRects.test.ts
+++ b/packages/folio/src/core/layout-bridge/selectionRects.test.ts
@@ -552,6 +552,149 @@ describe("selection rect geometry", () => {
     expect(rects.every((rect) => rect.y >= 0 && rect.y < 40)).toBe(true);
   });
 
+  test("paints a fixed-width sliver for an empty paragraph the selection spans (#836)", () => {
+    // A genuinely empty paragraph has a zero-length line (pmStart === pmEnd),
+    // so the run-span overlap (sliceFrom >= sliceTo) produced no rect and
+    // dragging across blank lines showed no highlight. Emit a caret-height
+    // sliver instead.
+    const block: ParagraphBlock = {
+      kind: "paragraph",
+      id: "empty",
+      pmStart: 10,
+      pmEnd: 12,
+      runs: [{ kind: "text", text: "" }],
+    };
+    const measures: Measure[] = [
+      {
+        kind: "paragraph",
+        lines: [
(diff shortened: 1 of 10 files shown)


jan-kubica:
@codex review

jan-kubica:
Round 2 addressed: a zero-length empty-paragraph line now overlaps the selection inclusive of the endpoints, so a drag that starts or ends exactly at an empty paragraph slivers it (a collapsed caret still draws none). Endpoint test added. The header/footer clip finding was already handled in round 1 (clipping is gated on the content fitting the shrunk box).

jan-kubica:
@codex review

PR #755, opened 2026-06-18
fix(folio): honor docDefaults, HTML auto-spacing, and footnote-anchor superscript

jan-kubica:
Ports three upstream `eigenpal/docx-editor` spacing/style fidelity fixes into folio. Independent of the sibling sync PRs (disjoint files).

### Changes
- **docDefaults precedence** (`eigenpal/docx-editor#909`): the built-in Normal fallback (8pt-after / 1.08 line) is synthesized only when the document defines neither a `Normal` style nor `w:docDefaults`. A document that ships its own `docDefaults` (common in generated files) keeps them, so style-less paragraphs (e.g. table cells) render at the document's compact height instead of the inflated template spacing.
- **HTML auto-spacing** (`eigenpal/docx-editor#823`): paragraphs with `w:beforeAutospacing`/`w:afterAutospacing` render Word's ~14px auto gap, overriding any explicit before/after — applied in both the paged layout (so pagination matches) and the live-edit DOM. Adds a named `AUTO_PARAGRAPH_SPACING_PX` constant.
- **Footnote-anchor superscript** (`eigenpal/docx-editor#845`): a footnote/endnote reference anchor renders superscript by default (Word's built-in FootnoteReference style), even when the source OOXML omits an explicit `vertAlign` (e.g. a bare `<w:footnoteReference/>`).

### Tests
- `styleResolver.test.ts` (docDefaults precedence, bare-document fallback), `formatToStyle-autospacing.test.ts`, `toFlowBlocks.test.ts` (auto-spacing pagination), `toFlowBlocks-run-marks.test.ts` (anchor superscript).

Full folio suite green.

### DIFF
--- a/packages/folio/src/core/docx/serializer/paragraphSerializer.ts
+++ b/packages/folio/src/core/docx/serializer/paragraphSerializer.ts
@@ -268,12 +268,16 @@ function serializeSpacing(formatting: ParagraphFormatting): string {
     attrs.push(`w:lineRule="${formatting.lineSpacingRule}"`);
   }
 
-  if (formatting.beforeAutospacing) {
-    attrs.push('w:beforeAutospacing="1"');
+  if (formatting.beforeAutospacing !== undefined) {
+    attrs.push(
+      `w:beforeAutospacing="${formatting.beforeAutospacing ? "1" : "0"}"`,
+    );
   }
 
-  if (formatting.afterAutospacing) {
-    attrs.push('w:afterAutospacing="1"');
+  if (formatting.afterAutospacing !== undefined) {
+    attrs.push(
+      `w:afterAutospacing="${formatting.afterAutospacing ? "1" : "0"}"`,
+    );
   }
 
   if (attrs.length === 0) {
--- a/packages/folio/src/core/docx/serializer/runSerializer.test.ts
+++ b/packages/folio/src/core/docx/serializer/runSerializer.test.ts
@@ -127,6 +127,19 @@ const ANY_DECIMAL_IN_EMU_ATTR =
   /(?:cx|cy|distT|distB|distL|distR|lIns|tIns|rIns|bIns)="-?\d+\.\d+"/u;
 const POSOFFSET_DECIMAL = /<wp:posOffset>-?\d+\.\d+<\/wp:posOffset>/u;
 
+describe("runSerializer vertical alignment", () => {
+  test("serializes explicit baseline vertical alignment", () => {
+    const xml = serializeRun({
+      type: "run",
+      formatting: { vertAlign: "baseline" },
+      content: [{ type: "footnoteRef", id: 1 }],
+    });
+
+    expect(xml).toContain('<w:vertAlign w:val="baseline"/>');
+    expect(xml).toContain('<w:footnoteReference w:id="1"/>');
+  });
+});
+
 describe("image EMU attributes are integer-only (issue #417)", () => {
   test("inline image with float dimensions serializes integer cx/cy/effectExtent", () => {
     const xml = serializeRun(FLOAT_INLINE_IMAGE);
--- a/packages/folio/src/core/docx/serializer/runSerializer.ts
+++ b/packages/folio/src/core/docx/serializer/runSerializer.ts
@@ -399,7 +399,7 @@ export function serializeTextFormatting(
   }
 
   // Vertical alignment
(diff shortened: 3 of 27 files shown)


jan-kubica:
Addressed the review findings:

- **Empty-but-present `docDefaults`** (codex): `parseDocDefaults` now returns a defined (possibly empty) object whenever the `<w:docDefaults>` element exists, so the resolver treats an explicitly-empty default as zero spacing rather than synthesizing the built-in template. Parse- and resolver-level tests added.
- **Empty-paragraph auto spacing** (codex): auto spacing is marked `spacingExplicit` so an imported empty auto-spaced paragraph keeps the ~14px gap instead of being collapsed. Test added.
- **Footnote-anchor superscript vs explicit subscript** (gemini): resolved after the mark loop so an explicit subscript wins regardless of mark order. Test added.

On the auto-spacing-vs-explicit guard (gemini, `toFlowBlocks` / `ParagraphExtension`): keeping auto spacing authoritative here is deliberate. Word writes `w:before="0"` alongside `w:beforeAutospacing="1"` as a fallback for consumers that don't support auto spacing, so honoring the explicit value would render `0px` instead of the ~14px auto gap for the common import — i.e. the guard would regress the very case this change targets. Respecting an in-editor spacing edit over imported auto spacing is a separate concern (it would clear the auto-spacing flag on edit), out of scope for this import-fidelity change.

jan-kubica:
@codex review

jan-kubica:
Round 2:

- **Stale auto-spacing on edit:** fixed. Auto-spacing now applies only while the paragraph's `spaceBefore`/`spaceAfter` still equals the imported value (`_originalFormatting.space*`); once the editor sets a different explicit value, that edit wins over the auto-spacing flag, in both the pagination path (`toFlowBlocks`) and the live-edit DOM (`ParagraphExtension`). This keeps Word's import behaviour intact (Word writes `before="0"` alongside `beforeAutospacing`, so auto wins while unedited) without ignoring a real edit. Test added.
- **Explicit `vertAlign="baseline"` on a note reference:** folio's PM model only carries super/subscript as marks; `toProseDoc` emits no mark for `baseline`, so a run that explicitly requested baseline is indistinguishable from one with no vertical alignment by the time it reaches the layout bridge. Suppressing the footnote-anchor default for that case would need a separate vertAlign-override signal plumbed end-to-end (parse to PM to flow), which is out of scope here; tracking it for the model-plumbing pass.

jan-kubica:
@codex review

jan-kubica:
Round 3 addressed: the auto-spacing "unchanged since import" gate now treats an import with no DIRECT spacing (`_originalFormatting.space* === undefined`) as unchanged, so a style/default-supplied value in the PM attrs (or a raw `null`) no longer looks like a user edit and the auto gap still wins — in both `toFlowBlocks` and `ParagraphExtension`. The `null` vs `undefined` normalization is subsumed by that absent-original clause. Test added for the style-supplied-spacing case.

jan-kubica:
@codex review

jan-kubica:
Round 5 addressed: `setSpaceBefore`/`setSpaceAfter` now clear the matching imported auto-spacing flag (`_originalFormatting.before/afterAutospacing`) for the edited edge. So once the user sets an explicit value, the auto gap no longer overrides it in the paged layout or the live DOM, and the saved DOCX drops `w:before/afterAutospacing` for that edge (the serializer reads it from `_originalFormatting`). The unedited edge and the import-fidelity behaviour are untouched. Command test added.

jan-kubica:
@codex review

jan-kubica:
Round 6 addressed: `setParagraphSpacing` now also copies the new value into `_originalFormatting` (the snapshot `fromProseDoc` serializes from), not just the auto flag. So an explicit spacing edit on an imported paragraph saves the edited value instead of the old imported one (often `0`). Test extended to assert the value is mirrored into the snapshot.

jan-kubica:
@codex review

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

PR #756, opened 2026-06-18
fix(folio): preserve per-run RTL, w:rStyle, and collapsed PAGE-field formatting on round-trip

jan-kubica:
Ports three upstream `eigenpal/docx-editor` round-trip preservation fixes into folio. Independent of the sibling sync PRs (disjoint files). Adds one field to `@stll/docx-core`.

### Changes
- **Per-run RTL in mark helpers** (`eigenpal/docx-editor#822`, #806): the conversion path handled `rtl`, but the live-edit / clipboard / keymap helpers (`marksToTextFormatting` / `textFormattingToMarks`) had no rtl branch, so an Arabic/Hebrew run lost its direction the moment it was re-marked in the editor. Added the branch (the negative override still rides `runFormattingOverride`).
- **`w:rStyle` run character-style** (`eigenpal/docx-editor#833`): a run's named character-style reference was dropped on the PM save round-trip (its formatting is flattened onto direct marks). A new inert `runStyle` mark carries the styleId through, so `<w:rStyle>` is re-emitted. Follows folio's existing run-shading mark pattern (schema attrs + attrs reader + validation case + StarterKit registration + conversion both ways).
- **Collapsed PAGE-field formatting** (`eigenpal/docx-editor#909`): a complex field with no separate result run (e.g. a footer `PAGE` number whose `w:rPr` lives on the field run) dropped its size/color. The formatting is now captured onto `ComplexField.formatting` (new field in `@stll/docx-core`) and used as the render/serialize fallback.

### Tests
- `markUtils.test.ts` (rtl round-trips through the mark helpers), `run-rstyle-roundtrip.test.ts`, `paragraphParser.test.ts` (complex-field formatting capture).

Full folio suite green.

### DIFF
--- a/packages/docx-core/src/model/content.ts
+++ b/packages/docx-core/src/model/content.ts
@@ -309,6 +309,13 @@ export type ComplexField = {
   fieldCode: Run[];
   /** Display result runs */
   fieldResult: Run[];
+  /**
+   * Run formatting captured from the field's structural run(s) (`w:rPr` on the
+   * `begin`/code run). Used as the run-formatting fallback when the field has
+   * no separate result run to read it from, e.g. a footer `PAGE` number whose
+   * formatting lives on the field run (eigenpal/docx-editor#909).
+   */
+  formatting?: TextFormatting;
   /** Field is locked */
   fldLock?: boolean;
   /** Field is dirty */
--- a/packages/folio/src/components/DocxEditor.tsx
+++ b/packages/folio/src/components/DocxEditor.tsx
@@ -1224,7 +1224,7 @@ export function DocxEditor({
         for (const [rId, hf] of bag) {
           const view = editor.getHfView(rId);
           if (view) {
-            hf.content = proseDocToBlocks(view.state.doc);
+            hf.content = proseDocToBlocks(view.state.doc, doc.package.styles);
           }
         }
       };
--- a/packages/folio/src/components/hooks/useHeaderFooterEditor.ts
+++ b/packages/folio/src/components/hooks/useHeaderFooterEditor.ts
@@ -472,7 +472,10 @@ export const useHeaderFooterEditor = ({
         // brand-new Map so the previous Document referenced by every
         // earlier history entry stays untouched and undo can step
         // back to the pre-edit state.
-        const blocks: BlockContent[] = proseDocToBlocks(view.state.doc);
+        const blocks: BlockContent[] = proseDocToBlocks(
+          view.state.doc,
+          pkg.styles,
+        );
         const updated: HeaderFooter = {
           ...existing,
           type: hfEditPosition,
--- a/packages/folio/src/core/docx/paragraphParser.test.ts
+++ b/packages/folio/src/core/docx/paragraphParser.test.ts
@@ -394,3 +394,61 @@ describe("parseParagraph SDT content preservation", () => {
     expect(outer.content[0].type).toBe("inlineSdt");
   });
 });
+
+describe("parseParagraph complex field formatting (#909)", () => {
+  test("captures the field run's w:rPr when there is no separate result run", () => {
+    // Footer pattern: begin / instrText / separate / end all in a single run
+    // carrying the w:rPr, with no separate result run. The formatting must
+    // survive on the ComplexField so the rendered PAGE number keeps the
+    // footer's size and color instead of falling back to the default.
+    const paragraph = parseParagraphXml(`
+      <w:p xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
+        <w:r>
+          <w:rPr><w:color w:val="595959"/><w:sz w:val="14"/><w:szCs w:val="14"/></w:rPr>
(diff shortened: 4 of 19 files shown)


jan-kubica:
Addressed both review findings on the `runStyle` mark:

- **Required `styleId`** (gemini): removed the `null` default so the schema attr matches `RunStyleAttrs` (a required `string`) and `readRunStyleMarkAttrs`. The mark is only ever created with a styleId, and `parseDOM` returns `false` without the marker.
- **Edit-reversion via the preserved link** (codex): the link is now reconciled against the character style on save. `fromProseDoc` builds a style engine from the base document's styles and, for each run carrying a `runStyle` reference, drops the `styleId` when the run's direct formatting no longer matches the style (e.g. style-derived bold toggled off). So an untouched styled run keeps its `w:rStyle`, but an edited one no longer has the style silently re-applied in Word. Tests cover both the keep and drop paths.

The other two fixes in this PR (#822 RTL, #909-B PAGE field) are unchanged.

jan-kubica:
@codex review

jan-kubica:
Round 2 addressed: `toProseDoc` now flattens with the document's own styles by default, so an unedited run's marks always reflect the style and a no-op save no longer strips `w:rStyle` (the `:253` concern). The save-side divergence check now also covers `rtl`, non-toggle removals (underline/color/size/font), and runs nested in fields, inline SDTs, and tracked-change containers. Tests added for the unexpanded-keep and non-toggle-drop paths.

jan-kubica:
@codex review

jan-kubica:
Round 3 addressed:

- **Auto-colour styles (`fromProseDoc.ts:317`):** the colour divergence check now skips an auto colour (`styleFormatting.color && !styleFormatting.color.auto`), mirroring toProseDoc which emits no textColor mark for `w:color val="auto"`. An unedited run carrying such a style no longer reads as diverging, so its `w:rStyle` is kept.
- **Field run styles (`toProseDoc.ts:2437`):** `convertField` now threads the style resolver and flattens the field result's own character style (`getRunStyleOwnProperties`) before building marks, the same precedence as `convertRun`/`convertHyperlink`. A PAGE/REF field styled only by a `w:rStyle` now carries its style-derived marks, so the reconciliation keeps the link. Tests added for both.

jan-kubica:
@codex review

jan-kubica:
Round 4 addressed: `proseDocToBlocks` now takes the document styles and runs the same `reconcileRunStyleReferences` pass when they're provided. The header/footer save paths (`useHeaderFooterEditor` and `DocxEditor`'s flush-on-save) now pass `package.styles`, so a character-style edit inside a header or footer is reconciled the same way as the body and is no longer reverted on save. Without styles the call is unchanged (no reconciliation context). Test added covering both.

jan-kubica:
@codex review

jan-kubica:
Round 5 addressed structurally: `runDivergesFromStyle` no longer hand-lists properties. It now compares the marks `toProseDoc` would emit for the style against the marks the run carries, and treats the link as diverged when the run is missing a mark *type* the style contributes. This automatically covers every property that round-trips (highlight, shading, spacing, position, kerning, etc.), keeps overrides (same mark type, different value) as non-divergent, and correctly ignores values that emit no mark (auto colour, baseline) — so it stays in lockstep with `textFormattingToMarks` instead of a subset. Test added for a previously-missed property.

jan-kubica:
@codex review

jan-kubica:
Round 6 addressed:

- **Compound marks (`fromProseDoc.ts:294`):** the divergence check now compares mark *attributes*, not just type names. The run still matches only if its mark of the same type carries every sub-property the style sets; dropping one (e.g. `fontFamily.eastAsia` while keeping `ascii`/`hAnsi`) now diverges, while merely overriding a value stays non-divergent.
- **RTL empty paragraphs (`markUtils.ts` gate):** `rtl` is now included in `EmptyParagraphFormatExtension.hasNonFontDefaults`, so the caret entering an empty RTL-styled paragraph re-derives the `rtl` stored mark and typed text keeps `w:rtl` instead of going LTR.

Tests added for both.

jan-kubica:
@codex review

jan-kubica:
Round 7 addressed (both in the #909 field-formatting path; the #833 reconciliation is unchanged): the collapsed-field formatting fallback now applies only when `fieldResult.length === 0`, so a present-but-unformatted result run keeps inheriting paragraph defaults instead of adopting the field code's colour/size. And the parser treats an empty `<w:rPr/>` on the begin run as absent, so a later genuinely-formatted run is captured instead of being blocked by the empty object. Tests added for both.

jan-kubica:
@codex review

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

jan-kubica:
@codex review

CC on behalf of @jan-kubica

PR #757, opened 2026-06-18
fix(folio): delete spanned table rows and track paste-over-selection

jan-kubica:
Two upstream `eigenpal/docx-editor` editor-command fixes. Independent of the sibling sync PRs (touches only `TableExtension.ts` and `suggestionMode.ts`).

- **Delete spanned table rows** (`eigenpal/docx-editor#783`): `deleteRow` now removes every row a multi-cell selection covers (via `selectedRect`), deleting the contiguous span in a single step, and drops the whole table when the selection spans all rows. (Track-changes deletion of rows is not modelled, so only the plain path is ported.)
- **Track paste over a selection** (`eigenpal/docx-editor#784`): with track changes on, pasting over a non-empty text selection records the replaced text as a deletion and the pasted content as an insertion (matching typing over a selection), then collapses the selection after the paste, instead of silently destroying the replaced text. A collapsed cursor falls through to the existing default-paste path.

Tests added for both.

Note: the internal plugin is named `suggestionMode` (legacy naming inherited from the upstream fork); the user-facing feature is track changes.

### DIFF
--- a/packages/folio/src/core/prosemirror/extensions/nodes/TableExtension.test.ts
+++ b/packages/folio/src/core/prosemirror/extensions/nodes/TableExtension.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, test } from "bun:test";
 import type { Node as PMNode } from "prosemirror-model";
 import { EditorState, TextSelection } from "prosemirror-state";
 import type { Transaction } from "prosemirror-state";
+import { CellSelection } from "prosemirror-tables";
 
 import { schema, singletonManager } from "../../schema";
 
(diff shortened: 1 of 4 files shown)


jan-kubica:
@codex review

jan-kubica:
Round 2 addressed: `deleteRow` now removes rows via prosemirror-tables' `removeRow` (rowspan-aware) instead of a raw node-span delete, so a cell spanning into a deleted row has its rowspan adjusted and content preserved; a plain caret deletes only its own row even inside a vertically merged cell (a `CellSelection` still deletes its full span). Tracked-paste collapse uses `TextSelection.near` so a block-content paste leaves a valid selection. Rowspan-adjust test added.

jan-kubica:
@codex review

jan-kubica:
Round 3 addressed: the tracked-paste insertion now uses `tr.replaceRange` instead of a raw `tr.replace` at the inline point, so the slice's open sides are fitted into the surrounding content the way ProseMirror's normal paste does. Block clipboard content (a copied table or whole paragraphs) is now placed structurally and still gets the tracked-insertion marks, instead of failing or dropping nodes. Added a block-content paste test.

jan-kubica:
@codex review

jan-kubica:
Round 4 addressed: deleting every row of a table that is its parent's only child now replaces it with an empty paragraph instead of a bare delete, so the document/cell keeps a valid block rather than becoming an empty (schema-invalid) node. When the table has siblings it is still removed outright. Test added asserting the doc validates after dropping a sole table.

jan-kubica:
@codex review

PR #771, opened 2026-06-20
feat(table): row grouping 

jan-kubica:
Builds on the unified condition AST (#750) to add two table primitives.

## Group table rows by a property
Collapsible sections in the table and kanban layouts, driven by any
groupable property. Shared grouping logic lives in `kanban-view.logic.ts`;
the table renders sections via `grouped-table-layout.tsx`.

## One-click Document Type classifier
A single action creates a `Document Type` single-select `ai-model`
property (starter legal document types, dependent on the workspace's file
property) and runs extraction over it. Re-running an existing classifier
is available without recreating it.

## Spend discipline: scope reruns to documents that need them
A property-scoped workflow run (`propertyIds` without `entityIds`)
previously always set `forcePropertyIds`, so `prepareBatch` recomputed
**every** document — re-running the classifier after adding a few files
re-spent on the whole set.

- Adds an explicit `force` flag to the `workflow/start` contract (default
  off). Unforced scoped runs keep the per-entity `prepareBatch` gate, so
  they only compute documents still missing a value (e.g. newly added
  ones). `force: true` is reserved for a deliberate full re-run.
- The classifier's primary action is unforced ("classify what's new"); a
  separate menu item forces a full reclassification.
- `prepareBatch` tests lock the invariant: a fresh cell with a result is
  skipped by default, recomputed only when forced, and `locked` still
  wins over `force`.

## Verification
api typecheck, api lint, `workflow/utils` + `views-schema` + grouped-table
tests green locally; web typechecks via CI.


jan-kubica:
All automated review findings on the grouped-table work have been addressed; resolving the threads. Map of finding → fix commit:

**Selection & justifications** (grouped sections share one selection/store)
- Resolve row selection to the toolbar across sections — `08f4f83a0`
- Preserve other sections' selections on a section's select-all/clear — `17ac9e586`
- De-duplicate selection across multi-select sections — `4eed3f9ea`
- Sync AI-cell justifications per loaded page — `08f4f83a0`

**Paging & scroll**
- Page via the real scroll viewport, not the inline content wrapper — `08f4f83a0`, `e3a599c91`

**Counts ↔ rows consistency**
- Exclude folders/tasks from grouped rows and counts — `d97253fff`
- Guard the group-counts unnest against scalar values — `054449ba8`
- Validate select-only group-counts server-side (read boundary) — `55b47d413`
- Render a section for stale (out-of-options) select values — `e3752d541`
- Drop cached rows when a group empties — `b88c7ef2a`
- Hide per-group sums until a group is fully loaded — `b88c7ef2a`

**Grouping eligibility**
- Reject status / non-select property groupings for document tables — `17ac9e586`, `b88c7ef2a`
- Offer multi-select grouping in the table picker; keep kanban single-select — `55b47d413`, `4eed3f9ea`

**Cross-workspace integrity**
- Remap table `groupByPropertyId` on duplicate / template apply — `f2dbd64fd`
- Clear a table grouping when its property is deleted — `2f04859c5`
- Collision-proof grouped section keys — `611bb20ce`

**Bulk review**
- Undo reverts only the cells the operation changed (and its auto-lock) — `e3a599c91`

Findings on the reverted one-click classifier (`classify-documents.tsx`, `workflow-queue.ts`) are moot — that code was removed from this PR.

CC on behalf of @jan-kubica

PR #780, opened 2026-06-21
feat(i18n): right-to-left support 

### LINKED ISSUE #775: feat: right-to-left support
### Feature request

Proper support for right-to-left languages requires not only translation of individual elements, but also mirroring the UX, making sure directionally-aware icons can point both directions etc.

### Motivation

Propert support for RtL languages. 

### Relevant sources

Also note eg Arabic needs -at minimum- support for different numbering (Eastern Arabic), calendar (Hijri + start of week shift).  Ideally this may be bundled in to be easier to review for native speaker as a complete localization effort

jan-kubica:
Adds right-to-left layout and a full Arabic (`ar`) locale, plus the locale-formatting layer it needs.

- Text direction follows the active locale (`dir`); full Arabic UI translation (Modern Standard Arabic)
- Central locale-aware formatter; calendar (Gregorian default, Hijri opt-in) and numeral (Western/Eastern) preferences in the account menu
- Currency formatted in the correct minor unit (e.g. KWD 3 decimals, JPY 0); locale-derived week start; `<bdi>`-isolated case numbers and citations
- folio: Arabic-script font fallback and a paragraph text-direction toggle

The Arabic translation is machine-generated and should get a native legal-Arabic review before release.


jan-kubica:
Addressed the review feedback in `dcf675ca`:

- **Currency minor-unit (P1):** reverted. The expense and rate forms write amounts as ×100, so a currency-aware read divisor breaks the round-trip for non-2-decimal currencies (JPY/KWD). Correct support needs a coordinated read + write + export change and a data consideration, so it's tracked separately; the formatter stays on the stored 1/100 scale.
- **prices-panel hydration (high):** now formats via `useFormatter()` (provider-aware, respects pre-hydration English) threaded through the helpers, instead of the store-level formatter.
- **Heatmap week start (P2):** day columns and labels rotate to the locale's first weekday, so they align with the localized week range.
- **week.ts:** guard against a non-numeric `firstDay` from `Intl`.
- **CI lint:** named the fallback capture group, un-nested the formatting-locale ternary, and annotated the intentional week-start memo dependency.

jan-kubica:
Addressed the latest review (`961adf89`):

- **AI search locale (16-char cap):** SearchDialog now sends the base language to `search.refine`/`search.summary`, not the extended formatting locale.
- **week.ts:** falls back to the legacy `Intl.Locale#weekInfo` accessor when `getWeekInfo()` is absent, so the locale-aware week start isn't silently Monday on older engines.
- **Hijri × Gregorian grids:** the calendar month/day grid labels (`calendar-utils` + `date-picker-popover`) are pinned to `calendar: "gregory"`, so a Hijri preference no longer mislabels Gregorian grids; standalone dates still render in the chosen calendar, and numerals still localize.

jan-kubica:
Addressed (`8ce07ef`):

- **pt-BR crash (P1):** `buildFormattingLocale` no longer appends a detected region to a language that already encodes one (`pt-BR`) — it was building an invalid `pt-BR-BR` tag that `Intl.Locale` rejects, crashing the provider. Added a regression test over every language × region.
- **Organizer locale cap (P2):** `existing-file-organizer` now sends the base language to `organize-suggestions` (16-char cap), matching the SearchDialog fix.

I audited every locale→backend payload: `search.refine`/`summary`, `entities.organize-suggestions`, and chat (uncapped, already base) — those were the only ones.

jan-kubica:
Rebased onto current `main` (resolved the React-Compiler / `useEffect`-ban conflicts) and addressed the three new P2s:

- **Arabic transactional emails:** added `getEmailDirection`; the OTP/invitation/new-device `<Html>` wrappers now carry `dir="rtl"` for Arabic.
- **Folio document direction:** `.folio-root` resets to `direction: ltr`, so an RTL app UI no longer flips the base direction of Latin (non-bidi) DOCX paragraphs; per-paragraph bidi still renders RTL content.
- **Calendar weekends:** derived from `Intl` `getWeekInfo().weekend` via `getWeekendDays` (Friday/Saturday for `ar-SA`), threaded through the calendar helpers. Western locales are byte-identical.

jan-kubica:
The date picker duplicated the week-start logic and, unlike `week.ts`, didn't read the legacy `Intl.Locale#weekInfo` accessor, so it fell back to Monday on engines that expose it. Extracted a single `getLocaleWeekInfo`/`getFirstWeekday`/`getWeekendDays` into `@stll/ui/lib/week` (legacy-accessor-aware); the date picker and `apps/web`'s `week.ts` now resolve the week start and weekend from one source.

jan-kubica:
Addressed (`f3358cf`):

- **Table exports preserve formatting prefs:** added `extractFormattingLocale` (keeps the `-u-ca`/`-u-nu` extensions, validated via `Intl.Locale`, behind a shared `matchSupportedLang`). CSV/XLSX now format dates/numbers with the client's calendar/numbering settings instead of the bare base language. Email/view-name catalog lookups still use the base `SupportedLang`.
- **Duration input hint:** the editable hint now renders ASCII decimal hours so it round-trips through `parseDuration` (which only accepts ASCII digits + `.`); read-only timesheet totals stay localized.

jan-kubica:
Moved the LTR reset from `.folio-root` (which also wraps the translatable `FormattingBar`) to `.paged-editor` (the document canvas). The editor chrome — toolbar and menus — now follows the app direction under an RTL UI, while the document canvas keeps an LTR base direction so Latin (non-bidi) paragraphs render correctly.

PR #797, opened 2026-06-21
chore: enable vitest disabled test lint rule

jan-kubica:
Enables the Vitest no-disabled-tests lint rule for test files.\n\nValidation:\n- bun --bun oxlint -c oxlint.config.ts --deny vitest/no-disabled-tests --no-error-on-unmatched-pattern '**/*.{test,spec}.{ts,tsx,js,jsx}' '**/__tests__/**/*.{ts,tsx,js,jsx}'\n- pre-commit hook passed


jan-kubica:
@codex review

jan-kubica:
Closing after review: this Vitest rule does not currently cover Stella’s `bun:test` import pattern, so the guard would be silent for the tests it is meant to protect. A Bun-aware rule or custom guard would be a better direction.

PR #802, opened 2026-06-21
feat(filters): faceted value counts in the filter builder

jan-kubica:
When building a filter on a single/multi-select property, show the per-value entity count next to each option (Notion-style facets), so it's clear which values actually have rows before applying.

- New `POST /entities/:workspaceId/property-facets` read endpoint returns counts per value for a property under the current filters.
- `condition-select-values` renders the value picker with counts + the option colour swatch; wired into the filter builder and the toolbar filter chips.
- Counts respect the active filters and tenant scope.

Built on the merged `@stll/conditions` AST; no schema changes (read-only over existing fields).

### DIFF
--- a/apps/api/src/handlers/entities/read-property-facets.ts
+++ b/apps/api/src/handlers/entities/read-property-facets.ts
@@ -1,0 +1,110 @@
+import { Result } from "better-result";
+import { and, eq, sql } from "drizzle-orm";
+import { t } from "elysia";
+
+import { entities, fields } from "@/api/db/schema";
+import { createSafeHandler } from "@/api/lib/api-handlers";
+import type { HandlerConfig } from "@/api/lib/api-handlers";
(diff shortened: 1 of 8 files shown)


jan-kubica:
Addressed the review findings on this PR.

**Fixed**
- `read-property-facets`: the `LATERAL jsonb_array_elements_text(...)` is evaluated for every left row before the join's `ON` predicate, so scalar (single-select/text) values crashed it. The function argument is now guarded with a `CASE` that yields an empty array for non-arrays; scalars fall through to the scalar extraction branch.

**Deferred (minor follow-ups)**
- Threading `excludedKinds` into facet counts: counts are workspace-wide over entities that carry the property's field, and folders/tasks rarely hold such fields, so the practical skew is negligible. Can be threaded through if wanted.
- Truncated (top-50) facets showing a missing value as `0`: acceptable for the "top values" use case; facets with >50 distinct values are rare.

PR #803, opened 2026-06-21
feat(playbooks): per-document-type column bundles with auto-gating

jan-kubica:
## Summary

Adds **playbooks**: per-document-type bundles of AI columns. A playbook binds a Document Type property value to a set of columns; applying it materializes each column as a workspace property gated so it only fills for matching documents.

## How it works

- A playbook stores a `typePropertyId` + `typeValue` and a `bundle` of column definitions (name, content, tool).
- Applying a playbook (idempotent by column name) creates any missing properties in one bulk insert, then attaches a `propertyDependency` gating each column with `<typeProperty> eq <typeValue>` via `@stll/conditions`.
- Gating reuses the existing extraction-dependency mechanism: no new trigger engine, no classifier button. Columns simply stay empty for documents whose type does not match.

## Scope

- API: `playbooks` table + RLS, CRUD handlers (`create`, `read`, `update-by-id`, `delete-by-id`, `apply`), validation, permissions, audit-log entries, limits.
- Web: `PlaybooksManager` in the view toolbar, queries, i18n (en source; ar/pl/pt-BR/sk regenerated, untranslated grandfathered in the baseline).

## Notes

- The apply path bulk-inserts all new columns in a single statement inside the transaction (no per-column round-trip).
- Property creation respects `LIMITS.propertiesCount`.


jan-kubica:
Addressed the review findings on this PR.

**Fixed**
- `apply`: materialized properties are now matched to bundle columns by a stable per-column `sourceId` (new nullable `properties.playbook_source_id`) instead of by name. Re-applying updates the property in place (rename-safe), never reuses a property owned by another playbook or a manually-created column, and re-runs every applied column under the current gate.
- `apply`: the type gate is re-validated via `validateTypeProperty` (single-select + option membership) before applying.
- `apply`: the gate dependency upserts with `onConflictDoUpdate`, so changing a playbook's `typeValue` and re-applying updates existing gates.
- A duplicate-column-name guard was added at the create/update boundary and in the editor.

**Intentional**
- When a workspace workflow is already running, `startWorkflow` returns `already-running`; `apply` treats this as success because the columns are committed as `stale` and picked up by the running/next run, matching the fire-and-forget contract of the other callers.

**Not issues**
- `property.tool.type`: `tool` is `NOT NULL` and is always a manual/AI union with a `.type`, so there is no crash path.
- `read.ts` cursor: mirrors the existing `invoices`/`rates` keyset pattern; the cursor value and the comparison both `date_trunc` to milliseconds with an `id` tiebreaker.
- optional chaining on `content.options`: the value is narrowed to `single-select`, where `options` is a structural invariant.

**Other**
- `apply` no longer matches by name, so a bundle column named like the Document Type property can no longer self-gate, and re-applying an edited column syncs its name/content in place.
- Bundle `maxItems` is a coarse upper bound; the real constraint (existing workspace properties + new columns ≤ `propertiesCount`) is enforced at apply time, which is the only point where the existing count is known.

PR #817, opened 2026-06-22
chore: update dependencies

jan-kubica:
## Summary
- update non-quarantined Bun dependencies across workspace manifests
- include selected major bumps for Atlaskit drag-and-drop packages and Babel
- keep Drizzle prerelease pins unchanged

## Validation
- bun install
- bun run lint:ws
- git diff --check
- bun pm untrusted
- pre-push format and i18n checks passed before local typecheck was skipped


jan-kubica:
^ all false positives

PR #864, opened 2026-06-25
feat(scripts): flag i18n terms duplicated across feature namespaces

jan-kubica:
## What

Extends the existing `i18n-check` gate. It already flags feature keys whose value duplicates a `common.*` value (`duplicatesCommon`); this adds detection for the **same value held by two or more feature keys** that is *not* in `common.*` — the signal that a term should be hoisted into `common.*` so it is translated once and stays consistent.

On a violation the gate prints, e.g.:

```
= shared value: billing.week (also workspaces.views.calendar.week, workspaces.views.timeline.week) — consider moving the term to common.* and reusing it, or baseline it
```

## Grandfathering

Intentional duplicates are allowed via a new `duplicateValues` list in `i18n-check-baseline.json`, mirroring the existing `duplicatesCommon` mechanism. The current debt (340 keys) is seeded so the gate stays green, and `--write-baseline` regenerates it. New duplicates introduced after this lands fail the check until reused, hoisted, or explicitly baselined.

## Notes

- Trivially-identical values (brand tokens, placeholder-only, acronyms) and values already covered by `common.*` are excluded to avoid double-reporting.
- Baseline change is purely additive: `duplicatesCommon` and `identicalToSource` are unchanged.
- New unit tests cover the finder; the full `i18n:check` pipeline passes.

## Summary by CodeRabbit

* **New Features**
  * Added a shared, localised “Formula” label across the app, and updated the condition formula UI to use a consistent prompt/placeholder.

* **Bug Fixes**
  * Standardised common wording for version, import, regenerate, and “new matter” actions across multiple screens (including clause history/metadata and sidebar/breadcrumb labels).
  * Improved consistency of translated labels across languages, so formula and version-related text is more reliably displayed.

* **Tests**
  * Updated i18n check coverage for shared-value duplicate detection.


jan-kubica:
Bot review dispositions:

- **glossary-gen `nouns` validation** (codex P2 / coderabbit Major): fixed in 93aff778b — a present-but-non-array `nouns` now `panic`s instead of silently defaulting to `[]`, matching how `verbs`/`legalConcepts`/`ptBR` are validated.
- **template-category-sidebar.tsx:112 placeholder** (coderabbit): not a regression — the label and placeholder already shared the same key (`templates.categoryName`) before this PR; the consolidation only repointed both to `common.categoryName`, so the rendered copy is unchanged. Adding a distinct example-style placeholder is a separate UX enhancement, out of scope here.
- **index.tsx:408 newMatter migration** (coderabbit): line 408 renders `common.newMatter` ("New matter") with text identical to the pre-PR `workspaces.newMatter`. Lines 274/582 use `workspaces.createNewWorkspace` ("Create new matter"), a deliberately distinct label for the empty-state CTA and context menu — not the same string, so no loading/settled flip is introduced by this change.
- **lv.json bulk.promptPlaceholder** (coderabbit): pre-existing untranslated string, not touched by this PR. Tracking separately as translation debt rather than cherry-picking one string here.

CC on behalf of @jan-kubica

jan-kubica:
Restrict duplicate grouping to cross-namespace pairs (codex P2): fixed in 591907ea0. `findSharedValueDuplicates` now skips any value-group whose keys all share one top-level namespace, so same-feature repeats (e.g. `workspaces.views.calendar.week` + `workspaces.views.timeline.week`) no longer trigger the "hoist to common.*" rule. Added a unit test for the same-namespace case; baseline drops from 218 to 148.

CC on behalf of @jan-kubica

PR #882, opened 2026-06-26
chore(db): single-source the corpus backfill statement_timeout

jan-kubica:
## What

Single-sources the `statement_timeout` used by the corpus backfill writes.

The tsvector search projections (case-law + legislation) and the citation-authority recompute each raised `statement_timeout` to a duplicated `'15min'` literal across three files. This extracts a shared `CORPUS_BACKFILL_STATEMENT_TIMEOUT` constant so the values can't drift apart, and documents that the bound governs an actively executing statement — independent of the DB-level `idle_in_transaction_session_timeout`, which only applies to a transaction sitting idle between statements.

No behaviour change.

## Test plan

- [x] typecheck (api)
- [x] type-aware oxlint on the changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of search index and backfill updates for case law and legislation.
  * Reduced the risk of long-running database operations timing out during large index maintenance tasks.
  * Kept normal user-facing queries on the standard database timeout, helping avoid unintended side effects.

### DIFF
--- a/apps/api/src/handlers/case-law/citation-authority.ts
+++ b/apps/api/src/handlers/case-law/citation-authority.ts
@@ -3,7 +3,7 @@ import { sql } from "drizzle-orm";
 
 import { courtWeightSql } from "@/api/handlers/case-law/citation-score";
 import { redistributableCaseLawSourceSqlFor } from "@/api/handlers/case-law/redistribution";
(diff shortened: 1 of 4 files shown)


jan-kubica:
@coderabbitai review

CC on behalf of @jan-kubica

PR #925, opened 2026-06-30
fix(chat): suppress follow-up suggestions while the ask-user card is live

jan-kubica:
When the assistant renders an ask-user clarification card, that card already presents its own questions and submit button. The generic follow-up suggestion chips (and the Tab-to-ask editor hint) were also shown on the same turn, giving two competing "answer next" affordances.

Gate the suggested-prompts query on the last assistant message not carrying a `tool-ask-user` part. While the clarification card is the live message, the suggested-prompts fetch is disabled, so the chips render nothing and the editor hint clears; suggestions resume once a fresh assistant message arrives.

### DIFF
--- a/apps/web/src/components/chat/ask-user-card.tsx
+++ b/apps/web/src/components/chat/ask-user-card.tsx
@@ -24,6 +24,7 @@ import type {
 } from "@/components/chat/chat-ui-tools";
 import { EntityLink } from "@/components/chat/entity-link";
 import { rehypeAnonSpans } from "@/components/chat/rehype-anon-spans";
+import { useExternalSyncEffect } from "@/hooks/use-effect";
 
 type AskUserPart = ToolUIPart<Pick<ChatUITools, "ask-user">>;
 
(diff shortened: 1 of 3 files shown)


jan-kubica:
Addressed both review notes in cd6533a6d:

- Gate now checks `part.state !== "output-available"`, so suppression only applies to a *pending* ask-user card. Once the card is answered (and the assistant may continue on the same message), follow-up suggestions return.
- `lastMessage` is narrowed explicitly (`lastMessage !== undefined && lastMessage.role === "assistant"`) before accessing `.parts`.

jan-kubica:
Addressed in d38c356f5: `AskUserCard` now reports its local edit-mode up through `ChatThreadMessages` to the page (`onAskUserEditingChange`), and the suggestion gate suppresses chips + the Tab-to-ask hint while a card is awaiting input *or* being edited. The signal is cleared on cancel, rerun, and unmount.

jan-kubica:
Addressed in 02b752862: the gate now keys off the edit-mode set being non-empty (`askUserOwnsTurn = pending-on-last-message || editingAskUserToolCallIds.size > 0`), so editing any reopened card — including an earlier one with downstream replies — suppresses the chips and Tab-to-ask hint regardless of its position in the transcript.

PR #957, opened 2026-07-02
fix: Enter advances the jurisdiction, catalogue, and AI onboarding steps

jan-kubica:
## Summary

Only the organization step advanced on <kbd>Enter</kbd>, because it wraps its content in a form with a submit button. The jurisdiction, catalogue, and AI-provider steps used `type="button"` + `onClick` handlers with no surrounding form, so a keypress had nothing to trigger.

Each of these steps now:
- wraps its content in a `<Form>` (using `display:contents` where the existing flex layout must be preserved unchanged),
- switches the primary "continue" button to `type="submit"`,
- autofocuses the search input so a keypress has a focused control to submit from.

The AI-provider step's submit is phase-aware and only advances when the current phase is valid (a saved provider for `providers` → `models`, valid model selection for continue). The shared `ai-config-providers-editor` is untouched; its buttons remain `type="button"`, so entering/saving an API key never triggers the step's submit.

The invite step is intentionally left as-is (its Enter commits an email chip).

## Summary by CodeRabbit

* **New Features**
  * Onboarding steps now support submitting with the Enter key, making it easier to continue through the flow.
  * Search fields in the catalogue and jurisdiction selection can now auto-focus for faster input.

* **Bug Fixes**
  * Improved onboarding navigation so primary actions behave consistently through form submission.
  * The jurisdiction step now only allows continuing when a selection has been made.

### DIFF
--- a/apps/web/src/components/jurisdiction-picker.tsx
+++ b/apps/web/src/components/jurisdiction-picker.tsx
@@ -22,12 +22,14 @@ type JurisdictionPickerProps = {
   selected: readonly PracticeJurisdiction[];
   suggestedCountryCodes?: readonly CountryCode[];
   onChange: (jurisdictions: PracticeJurisdiction[]) => void;
+  autoFocus?: boolean;
 };
 
 export const JurisdictionPicker = ({
   selected,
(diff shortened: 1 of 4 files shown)


jan-kubica:
false positive, intended

PR #1158, opened 2026-07-13
fix: enforce AI provider request contracts

shanehobson:
## Summary

Fixes three provider-boundary failures and adds coverage at the layer that was missing:

- OpenAI no longer receives the invalid `prompt_cache_retention: "in-memory"` option. `prompt_cache_key` remains enabled; retention is omitted because no explicit value is portable across the offered model catalogue.
- OpenAI free-form map tools use TanStack's existing `strict: false` fallback instead of being mislabeled `strict: true` and rejecting the entire request.
- Anthropic distinguishes Stella's ordinary `web_search` function from Anthropic's provider-native web-search tool by metadata, not by a colliding public function name.

The two TanStack runtime fixes are carried as pinned Bun patches and protected by adapter-boundary regression tests, so they can be removed independently once upstream releases the fixes.

## Why existing tests missed this

The test suite validated provider-neutral JSON Schema and exercised chat E2E through `USE_MOCK_AI=true`. Neither path observed the final request produced by a concrete provider adapter. The invalid cache option was also only reachable with prompt caching enabled, while generation tests used the disabled-caching fixture.

## Provider contract architecture

The mandatory offline suite now serializes every registered Stella chat tool through the real final request path for all six supported TanStack providers:

- OpenAI Responses
- Google Gemini
- Anthropic Messages
- Amazon Bedrock Converse
- OpenRouter Responses
- Mistral

It fails if any tool disappears before the SDK boundary. OpenAI additionally gets a strict-schema invariant with an exact non-strict allowlist, and cache-enabled generation asserts the complete model options plus the absence of `prompt_cache_retention`.

The new scheduled/manual live canary complements the offline gate. On the trusted default branch it probes text generation, structured output, closed strict tools, free-form map tools, and cache-enabled generation using synthetic inputs, the catalogue's `fast` model, a 64-token cap, and 20-second per-probe timeouts. It never logs prompts, responses, headers, keys, provider bodies, or stacks. Each matrix job receives only its selected provider credential; the workflow has least-privilege permissions, pinned actions, lifecycle scripts disabled during install, and fails if no provider canary is configured.

This follows TanStack AI's useful final-SDK-payload and provider-feature-matrix patterns, while closing two upstream gaps: its real-provider tool scripts are manual rather than CI canaries, and its `aimock` matrix cannot exercise native Bedrock Converse.

## Verification

- `bun install --frozen-lockfile --force --ignore-scripts`: both pinned patches apply cleanly.
- Focused provider/cache suite: 37 passed, 0 failed, 485 assertions.
- API typecheck: passed.
- Type-aware API lint: 0 warnings, 0 errors.
- Formatting: passed.
- `actionlint .github/workflows/ai-provider-canary.yml`: passed.
- Staged secret scans: no leaks.
- Security review: no critical/high open dependency alerts; no untrusted-trigger or secret-exposure path in the workflow.

The full local verifier completed skill sync, workspace hygiene, policy evidence, i18n, affected-package lint, and formatting successfully. Its unchanged web typecheck was stopped due local runtime; required GitHub CI remains the merge gate.

## Follow-ups

- Upstream the OpenAI open-object strictness fix and Anthropic native-tool discriminator as separate TanStack AI PRs.
- Replace the carried patches when fixed TanStack releases are adopted.
- If Stella needs a retention guarantee rather than the provider default, model that as a per-model catalogue capability; current OpenAI model generations do not share one portable retention field/value.

## Summary by CodeRabbit

* **New Features**
  * Added automated daily and on-demand checks for configured AI provider capabilities.
  * Added a command to run provider capability checks locally or in CI.
  * Added coverage reporting to confirm configured providers were tested.

* **Bug Fixes**
  * Improved compatibility of tool schemas with strict OpenAI requests.
  * Corrected handling of Anthropic web-search tools.
  * Improved prompt caching configuration portability.

* **Tests**
  * Expanded coverage across supported AI providers, tool conversion, schemas, and prompt caching.

### DIFF
--- a/.github/workflows/ai-provider-canary.yml
+++ b/.github/workflows/ai-provider-canary.yml
@@ -1,0 +1,125 @@
+name: AI provider canary
+
+on:
+  schedule:
+    - cron: "17 3 * * *"
+  workflow_dispatch:
+    inputs:
+      provider:
+        description: Provider to probe
+        required: true
+        default: all
+        type: choice
+        options:
+          - all
+          - google
+          - openrouter
+          - openai
+          - anthropic
+          - bedrock
+          - mistral
+
+permissions: {}
+
+jobs:
+  provider-canary:
+    name: ${{ matrix.provider }} capability contract
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        provider: ${{ fromJSON(github.event_name == 'workflow_dispatch' && inputs.provider != 'all' && format('["{0}"]', inputs.provider) || '["google","openrouter","openai","anthropic","bedrock","mistral"]') }}
+    concurrency:
+      group: ai-provider-canary-${{ matrix.provider }}
+      cancel-in-progress: true
+    steps:
+      - name: Check canary credential
+        id: credential
+        env:
+          AI_CANARY_API_KEY: ${{ matrix.provider == 'google' && secrets.GOOGLE_GENERATIVE_AI_API_KEY || matrix.provider == 'openrouter' && secrets.OPENROUTER_API_KEY || matrix.provider == 'openai' && secrets.OPENAI_API_KEY || matrix.provider == 'anthropic' && secrets.ANTHROPIC_API_KEY || matrix.provider == 'bedrock' && secrets.BEDROCK_API_KEY || matrix.provider == 'mistral' && secrets.MISTRAL_API_KEY || '' }}
+          EVENT_NAME: ${{ github.event_name }}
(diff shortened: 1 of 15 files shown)


jan-kubica:
@shanehobson thank you very much for the PR ! This has slipped CI checks and tests which should be impossible to happen with such a major issue. Root cause is partially in an upstream library (will open PRs) but I have still added some extra tests into this PR so it can merge asap and the planned release into prod (containing tanstack-ai migration and a lot of other changes) is unblocked

thank you again

jan-kubica:
Two upstream fixes, based on work of @shanehobson 
https://github.com/TanStack/ai/pull/932
https://github.com/TanStack/ai/pull/933

PR #1165, opened 2026-07-14
fix: preserve boolean types in provider schemas

jan-kubica:
## Summary

- retain the inferred boolean type when provider projection drops an unsupported boolean const constraint
- cover the type-less boolean literal shape emitted by the Valibot tool-schema path

## Why

Boolean literals can be serialized as a const without an explicit type. Dropping the exact constraint previously projected that schema to an unconstrained object, allowing providers to return a string where Stella expects a boolean.

## Testing

- bun test apps/api/src/lib/provider-safe-json-schema.test.ts
- oxfmt check on changed files
- type-aware oxlint on changed files
- git diff --check


jan-kubica:
folded into another PR

PR #1174, opened 2026-07-14
fix: normalize OpenAI strict tool inputs at the adapter boundary

shanehobson:
## What was broken

OpenAI strict tool schemas represent an omitted optional property as a required nullable property. The model can therefore return `null`, but TanStack AI 0.40.0 validated that provider-shaped input against Stella's original schema, where an optional property accepts absence rather than `null`. A conforming tool call could fail validation and abort the chat turn.

## Architecture

The provider adapter owns both sides of this representation change:

1. While producing the strict OpenAI schema, it records exactly which properties it widened with `null`.
2. When a tool call returns, it removes `null` only at those recorded paths before Stella validates or executes the tool.
3. Properties that were genuinely nullable in the original schema remain `null`.

Stella therefore keeps its domain contract unchanged: optional still means absent. We do not broaden every tool schema or normalize nulls independently in handlers.

The runtime fix is committed through Bun's exact-version `patchedDependencies` entry for `@tanstack/openai-base@0.9.7`, alongside the existing Stella patch for that package. It mirrors the behavior in TanStack AI PR [#939](https://github.com/TanStack/ai/pull/939). Once that fix is released and Stella upgrades, this patch section should be removed.

## Narrow compatibility guards

- The ask-user and spawn-subagents UI paths tolerate historical persisted tool arguments containing `null`.
- The scheduled provider canary now asks a tool call to omit an optional input and verifies the executed input remains absent. This catches removing the patch before adopting an upstream release containing the fix.
- A focused adapter regression test proves the installed package removes provider-added nulls while preserving a genuinely nullable field.

## Verification

- Forced `bun install --force --frozen-lockfile` succeeds, proving the committed dependency patch applies from a clean install.
- `tool-schema.test.ts`: 27 passed.
- Ask-user and canary configuration/coverage tests: 10 passed.
- Full local lint/typecheck were intentionally left to hosted CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of optional tool inputs, including cases where providers return `null` or omit fields.
  * Ask-user prompts now render as free-text when no selectable options are available, including historical `null` arguments.
  * Tightened canary round-trip validation to detect when an optional tool argument is incorrectly included.

* **Tests**
  * Added coverage for streamed tool-call chunking with nullable fields and for strict-to-nonstrict conversion fallback.
  * Added rendering and provider validation tests to prevent regressions.

### DIFF
--- a/apps/api/scripts/ai-provider-canary.ts
+++ b/apps/api/scripts/ai-provider-canary.ts
@@ -41,13 +41,14 @@ const TOOL_ROUND_TRIP_COUNT = 7;
 const TOOL_ROUND_TRIP_RESULT = "stella-tool-round-trip-ok";
 const TOOL_ROUND_TRIP_PROMPT =
   `Call ${TOOL_ROUND_TRIP_NAME} exactly once with value "${TOOL_ROUND_TRIP_VALUE}" ` +
-  `and count ${TOOL_ROUND_TRIP_COUNT}. Then reply with only the confirmation ` +
-  "value returned by the tool.";
+  `and count ${TOOL_ROUND_TRIP_COUNT}. Do not include optionalNote. Then reply ` +
+  "with only the confirmation value returned by the tool.";
 
 const SAFE_CANARY_ERROR_MESSAGES = new Set([
   "Canary resolved an unexpected provider model.",
   "Provider did not execute the canary tool exactly once.",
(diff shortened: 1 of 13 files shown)


jan-kubica:
@coderabbitai review

CC on behalf of @jan-kubica

PR #1283, opened 2026-07-20
fix(web): bound auth requests so a stalled connection cannot hang the shell

jan-kubica:
## Problem

The `require-fetch-timeout` oxlint rule forbids raw `fetch()` under `apps/web/src`, but it only reaches first-party call sites. The auth client does its fetching inside its own transport, so every `authClient` call was exempt from that invariant and carried no abort source of its own.

That matters on the boot path. `_protected`'s `beforeLoad` awaits the role query through that transport, and `DefaultPendingComponent` has no timeout and no error path, so a connection that stalls rather than fails (wedged HTTP/2 stream after sleep or a network switch) parks the route in a full-screen loader indefinitely, with nothing logged and no way out but a reload.

## Fix

Route the client through the existing `fetchWithTimeout` helper, and opt the two boot queries out of retries.

Two details that are load-bearing:

- **Bounding the request, not racing the promise.** An abort *settles* the query, so shell chrome still mounts against a resolved cache rather than an in-flight one. That preserves the ordering `beforeLoad`'s comment depends on (a role fetch resolving mid-mount triggers a React not-yet-mounted warning that route-smoke treats as a failure). The budget sits below `CRITICAL_QUERY_TIMEOUT_MS` so the transport settles first and that outer race stays a backstop.
- **Retries had to go with it.** The QueryClient sets no `retry`, so the TanStack default of 3 attempts plus exponential backoff applies. A per-attempt timeout alone would have stacked the budget four times over and merely slowed the hang down instead of removing it. Retries stay on for everything downstream of boot.

## Behaviour on timeout

| Path | Result |
|---|---|
| Session (`loadAuthContext`) | Settles as an error → existing catch → redirect to `/auth` |
| Role prefetch (`beforeLoad`) | Settles → prefetch swallows → chrome mounts without role-gated affordances |
| Post-boot queries | Unchanged |

Worst case on the boot path is now one request budget instead of unbounded.

## Follow-up worth a decision

The role-failure path degrades silently — the sidebar simply lacks role-gated affordances, with only an analytics event. Given #1260 in this release specifically converted analytics-only failures into surfaced ones, a toast may belong here too. Left out of this PR to keep the timeout change reviewable on its own.

### DIFF
--- a/apps/web/src/lib/auth.ts
+++ b/apps/web/src/lib/auth.ts
@@ -15,11 +15,34 @@ import { stellaToast } from "@stll/ui/components/toast";
 
 import { env } from "@/env";
 import { getTranslator, useI18nStore } from "@/i18n/i18n-store";
+import { fetchWithTimeout } from "@/lib/fetch";
 import { getSignedOauthQueryFromHash } from "@/lib/oauth-provider";
 import { createSecretTokenBoundary } from "@/lib/secret-token";
 import type { SecretToken } from "@/lib/secret-token";
 
 export const HTTP_TOO_MANY_REQUESTS = 429;
+
+/**
+ * Stall budget for every auth request.
+ *
+ * `require-fetch-timeout` only reaches first-party `fetch()` calls, so the
+ * auth client's own transport is the one hole in that invariant: a wedged
+ * connection (laptop sleep, network switch) leaves its promise pending
+ * forever. `_protected.tsx`'s `beforeLoad` awaits the role query through
+ * that transport, so an unbounded request there parks the route in its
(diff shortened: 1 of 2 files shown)


jan-kubica:
Reverted the `async`/`await` simplification suggested above — it fails CI here:

```
apps/web/src/lib/auth.ts:128:22: error typescript(promise-function-async):
Functions that return promises must be async.
```

`promise-function-async` runs type-aware with `--deny-warnings` in this repo, so the bare arrow form is rejected even though it is behaviourally equivalent. Added a comment at the call site so the simplification is not retried.

PR #1287, opened 2026-07-20
feat(scripts): ratchet capabilities that lose their schema to the byte cap

jan-kubica:
## Problem

`capInputSchema` (`apps/api/scripts/lib/capability-catalog.ts`) drops any capability input schema larger than `MAX_CAPABILITY_SCHEMA_BYTES` (64 KB) and marks the entry `inputSchemaTruncated`.

The consequence is not obvious from that code: a truncated capability loses its **typed CLI flags** and its **MCP input schema**. The command degrades to an opaque `--input` JSON blob, and `--help` cannot state the shape, so callers — humans and agents alike — have to already know it.

Nothing fails when this happens. There is no warning at export time and no check in CI, so the untyped surface can grow without anyone noticing.

## Current state

277 of 280 capabilities have usable schemas. Three do not:

- `views.create`
- `views.update`
- `view-templates.create`

## Guard

New `capability-schemas-truncated` ratchet metric, baseline **6** (three capabilities across both committed catalog mirrors — including both also catches drift between them). A fourth capability crossing the cap now fails CI; fixing one of the three ratchets the baseline down.

The metric counts committed artifacts rather than re-running the exporter, keeping the scan cheap and deterministic in line with the other metrics.

## Verification

Self-test passes, and covers the shapes the counter must **not** count (explicit `false`, and an entry omitting the field entirely):

```
$ bun scripts/ratchet.ts --self-test
ratchet --self-test: PASS
```

The guard was proven to fire, not just to pass — marking a fourth capability truncated produces:

```
ratchet --check: metric(s) rose above baseline:
  capability-schemas-truncated: 6 -> 7 (+1)
```

and reverting returns it to `OK. 13 metric(s) at or below baseline.`

## Not in scope

Fixing the three oversized schemas. This PR stops the bleeding; shrinking those schemas (or raising the cap deliberately) is separate work.

## Summary by CodeRabbit

* **Chores**
  * Added monitoring for truncated capability schemas in generated catalogues.
  * Updated baseline tracking to record six currently detected truncated schemas.
  * Expanded validation checks to ensure truncation counts remain accurate.

### DIFF
--- a/scripts/ratchet-baseline.json
+++ b/scripts/ratchet-baseline.json
@@ -756,6 +756,13 @@
       "apps/web/src/stores/config-store.ts": 2
     }
   },
+  "capability-schemas-truncated": {
+    "count": 6,
+    "files": {
+      "apps/api/src/mcp/generated/capability-catalog.json": 3,
+      "packages/cli/src/generated/capability-catalog.json": 3
+    }
+  },
   "cross-feature-imports": {
     "count": 0,
     "files": {}
--- a/scripts/ratchet.ts
+++ b/scripts/ratchet.ts
@@ -1181,7 +1181,8 @@ const SELF_TEST_CAPABILITY_CATALOG = `${JSON.stringify([
   { id: "gamma.update" },
   { id: "delta.delete", inputSchemaTruncated: true },
 ])}\n`;
-const EXPECTED_TRUNCATED_CAPABILITY_SCHEMAS = 2;
+// Two truncated entries in each of the two mirror fixtures.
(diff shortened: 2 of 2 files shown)


jan-kubica:
Good catch — applied. The self-test only wrote the CLI mirror, so the `apps/api/src/mcp/generated/capability-catalog.json` glob went unexercised, which is a real gap given the metric exists to keep the two artifacts in step.

Both fixtures are now written and the expectation is 4. Confirmed the assertion is not vacuous: setting it to 3 fails with `capability-schemas-truncated counted 4, expected 3`.

PR #1288, opened 2026-07-20
fix(cli): reach the manual-paste fallback when no browser can be opened

jan-kubica:
## Problem

`awaitAuthorizationCode` discarded the result of `openInBrowser`:

```ts
void openInBrowser(authorizeUrl);
```

`openInBrowser` spawns with `stdio: "ignore"` and correctly returns `false` when the launch fails, but nothing read it. On a machine that binds a loopback listener yet has no usable browser — headless server, SSH session, container, agent sandbox — the CLI then waited out the full `LOGIN_TIMEOUT_MS` (5 minutes) on a redirect that could never arrive before offering the manual paste that was available the whole time.

## Fix

Await the launch and divert to `awaitManualCode` immediately when it fails.

Closing the listener on that path is required rather than tidiness: it is an open `http.Server` with no `unref()`, so leaving it bound keeps the event loop alive and the CLI never exits, even after the pasted code succeeds. Verified with a minimal repro — a bound `createServer` with nothing else pending does not exit.

## Result

Headless sign-in is possible today: the URL is printed, the user completes it in any browser, and pastes the code back. That is not a substitute for a real non-interactive grant (see below), but it removes a five-minute dead wait on the path that already existed.

## Context: this is not the device-flow change

An attempt to enable better-auth's `deviceAuthorization` plugin was abandoned because the premise does not hold, verified against the installed dists:

- `/device/token` returns `access_token: session.token` — an opaque better-auth session token, with no `org_id`, no `aud`/`iss`, and no refresh token.
- `apps/api/src/mcp/auth.ts` verifies bearer tokens against JWKS with audience + issuer and hard-requires `payload.org_id`.
- `@better-auth/oauth-provider` supports `authorization_code` and `client_credentials` only — **no `device_code` grant** (zero occurrences in its dist).

So the plugin cannot produce a token the MCP surface accepts. Making it work would mean either loosening MCP to accept session tokens (dropping audience binding, `org_id`, and scopes — a security regression) or building a custom bridge that uses the device code purely as transport for the authorization code. Neither belongs in this PR.

For genuinely unattended clients, `client_credentials` is already supported by the OAuth provider and is the more promising direction. Tracked separately.

### DIFF
--- a/packages/cli/src/auth/browser-open.ts
+++ b/packages/cli/src/auth/browser-open.ts
@@ -18,10 +18,28 @@ export const openCommandFor = (
   return ["xdg-open"];
 };
 
-export const openInBrowser = async (url: string): Promise<boolean> => {
+/**
+ * Outcome of a launch attempt.
+ *
+ * Deliberately three states rather than a boolean: "the opener did not settle
+ * within the timeout" is NOT the same as "there is no browser". A slow desktop
+ * portal or an `xdg-open` that only exits once the browser closes both land in
+ * `unknown`, and the browser is very likely opening anyway — so callers must
+ * keep waiting on the loopback redirect there, and may only fall back to a
+ * manual paste on a definite `failed`.
+ */
+export type BrowserLaunch = {
+  readonly status: "opened" | "failed" | "unknown";
+};
+
+const LAUNCH_OPENED: BrowserLaunch = { status: "opened" };
+const LAUNCH_FAILED: BrowserLaunch = { status: "failed" };
+const LAUNCH_UNKNOWN: BrowserLaunch = { status: "unknown" };
+
+export const openInBrowser = async (url: string): Promise<BrowserLaunch> => {
   const command = openCommandFor(process.platform);
   if (!command) {
-    return false;
+    return LAUNCH_FAILED;
   }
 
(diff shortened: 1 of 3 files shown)


jan-kubica:
Both bots are right, and this was a real regression — thanks.

`openInBrowser` returned a bare `boolean` for **four** distinct outcomes: unsupported platform, spawn error, non-zero exit, and *the opener not settling within its 5s timeout*. Only the first three mean "no browser". A slow desktop portal, or an `xdg-open` that lives as long as the browser it spawned, lands in the fourth while the browser is opening perfectly normally — and my change closed the listener and dropped to manual paste, abandoning a redirect that was still on its way.

Fixed by making the outcome a three-state `BrowserLaunch` (`opened` | `failed` | `unknown`) rather than a boolean. Only `failed` diverts; `unknown` falls through to the loopback wait, which already has its own timeout fallback. Collapsing these into a boolean is precisely what made the regression possible, so the union is the structural fix rather than a patched condition.

Added a regression test for the `unknown` path, and confirmed it is not vacuous: relaxing the guard to `launch.status !== "opened"` fails it with a 5s timeout, and restoring passes. Full CLI suite 353 pass / 0 fail, type-aware lint clean.

PR #1289, opened 2026-07-20
fix(api): give every capability a kebab-case id

jan-kubica:
## Root cause

`deriveCapabilityId` builds an id from the handler's repo-relative path, then appends the raw TypeScript export identifier when the endpoint is a named export rather than the file's default:

```ts
return exportName === undefined ? base : `${base}.${exportName}`;
```

Export identifiers are camelCase by definition, so any handler file holding more than one capability endpoint leaked an internal function name into a **public** id. Exactly three files did this. The other 273 capabilities are single-endpoint files whose ids derive purely from the path.

Clean three-segment ids such as `case-law.matter-links.create` come from nested *directories*, not export names — that is the shape this aligns to.

## Renames

| old | new |
|---|---|
| `entities.read-summaries.readEntitySummariesCount` | `entities.read-summaries-count` |
| `workspaces.anonymization-allowlist.createWorkspaceAnonymizationAllowlistEntry` | `workspaces.anonymization-allowlist.create` |
| `workspaces.anonymization-allowlist.deleteWorkspaceAnonymizationAllowlistEntry` | `workspaces.anonymization-allowlist.delete` |
| `workspaces.anonymization-allowlist.readWorkspaceAnonymizationAllowlist` | `workspaces.anonymization-allowlist.read` |
| `workspaces.anonymization-terms.createWorkspaceAnonymizationTerms` | `workspaces.anonymization-terms.create` |
| `workspaces.anonymization-terms.deleteWorkspaceAnonymizationTerm` | `workspaces.anonymization-terms.delete` |
| `workspaces.anonymization-terms.readWorkspaceAnonymizationTerms` | `workspaces.anonymization-terms.read` |

Each named-export endpoint moved into its own kebab-case file as the default export. HTTP routes are byte-identical; only ids move. Catalog count unchanged at 280.

These seven previously had **no CLI command at all**, so this is a net +6 commands rather than a rename of a reachable surface.

## Structural guard (three layers)

1. `isWellFormedCapabilityId` + `CAPABILITY_ID_SEGMENT_PATTERN` in the catalog lib; the exporter validates every derived id and aborts generation and `--check`, failing CI.
2. The dispatch serializer reuses the same pattern, so a camelCase segment cannot reach the generated `import()` map even if it somehow passed layer 1.
3. An artifact test asserts every id in the committed snapshot is dotted kebab-case, catching a hand-edited or stale artifact independently of the exporter.

## Verification

Independently re-verified rather than taken on trust:

- Guard fires: tampering a committed id to `entities.readSomethingBad` fails the artifact test; reverting returns it to green.
- `jq '.[].id' | rg '\.[a-z0-9-]*[A-Z]'` → zero matches.
- Both catalog mirrors byte-identical after regeneration.
- All seven appear in the CLI tree, and all are present in the MCP dispatch map — both surfaces covered.
- `packages/cli` suite: 352 pass, 0 fail. `bun scripts/ratchet.ts --check`: OK.

## BREAKING CHANGE

These are public identifiers. Callers passing an old id to MCP `invoke_capability`, or scripting against a pinned catalog snapshot, will break. No deprecation alias is included: the old ids had no CLI surface, and an alias would reintroduce the malformed shape the guard now forbids. If a grace period is wanted, the right place is an old→new map at the `invoke_capability` boundary, not in the catalog.

## Note for review

`scripts/ratchet-baseline.json` still lists two now-deleted file paths in its per-file `lint-suppression-directives` map. Totals are unchanged so `--check` passes; the baseline was left untouched rather than reseeded, since reseeding needs justification.

## Summary by CodeRabbit

- **New Features**
  - Added workspace anonymisation allowlist and blacklist-term actions for reading, creating, and deleting entries.
  - Added an endpoint to return the total number of entities in a workspace.

- **Bug Fixes**
  - Improved capability catalogue validation to reject malformed identifiers before publication.
  - Updated capability and command mappings to use consistent lowercase kebab-case naming.

- **Documentation**
  - Updated capability coverage documentation with the revised identifiers and command paths.


jan-kubica:
Applied. Confirmed the spread was not load-bearing before removing it: `filter()` already returns a fresh array, so `sort()` mutates that copy rather than the `readonly string[]` input.

Verified rather than assumed — calling it with `["zeta.badId", "alpha.good", "beta.AlsoBad"]` returns `["beta.AlsoBad", "zeta.badId"]` and leaves the input array unchanged. Catalog tests: 104 pass.

PR #1290, opened 2026-07-20
fix(mcp): repair the better-auth doc source and surface failed sources

jan-kubica:
## Problem

`BetterAuth` was configured as `https://www.better-auth.com/llms.txt`. That host 307s to the apex:

```
HTTP/2 307
location: https://better-auth.com/llms.txt
```

`fetchAllowedUrl` follows redirects but re-runs `validateAllowedUrl` on **every hop**, against an allowlist derived from the `SOURCES` hostnames. `www.better-auth.com` was allowlisted; the redirect target `better-auth.com` was not. Reproduced directly:

```
OLD CONFIG FAILED: Blocked: better-auth.com is not a configured doc source
```

So the source has been indexing **nothing** since it was added.

## Fix

- Point the source at the canonical apex and alias the `www` host, matching the existing `bun.sh` → `bun.com` precedent for the same apex/www split.
- Name any sources that failed to fetch in the search response, on both the empty-results and results-returned paths, so a broken source is visible rather than inferred from suspiciously empty output.

## Verification

```
fetched bytes: 22447
mentions api-key plugin: true
```

The old configuration fails as diagnosed; the new one fetches and contains the content that was missing. MCP tests: 5 pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved documentation search reliability when the documentation site redirects between supported hostnames.
  * Search results now clearly indicate when one or more documentation sources could not be reached.
  * Added fetch failure details to empty search responses and returned results where relevant.

### DIFF
--- a/.claude/mcp/server.ts
+++ b/.claude/mcp/server.ts
@@ -628,7 +628,7 @@ server.tool(
       );
       const failureNote =
         failedSourceNames.length > 0
-          ? `\n\nNote: could not fetch ${failedSourceNames.join(", ")}; results exclude ${failedSourceNames.length === 1 ? "it" : "them"}.`
+          ? `Note: could not fetch ${failedSourceNames.join(", ")}; results exclude ${failedSourceNames.length === 1 ? "it" : "them"}.`
           : "";
 
       for (const { value } of successfulSources) {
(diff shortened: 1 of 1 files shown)


jan-kubica:
Both bots are right — appending the note to `JSON.stringify` made the results block unparseable for any caller treating tool output as JSON.

The note now rides as its own content block, so the results block stays valid JSON. Verified: `JSON.parse` on the first block succeeds, and the warning arrives as a second block only when a source actually failed. The empty-results path still concatenates, which is fine there since it is already prose, not JSON.

MCP tests: 5 pass.

PR #1292, opened 2026-07-20
feat(api): carry tool descriptions in the capability catalog

jan-kubica:
Stacked on #1291.

## Problem

Two definition layers described the same handlers:

- **Capability catalog** (280 entries, generated). Fields were exactly `access, destructive, handlerKind, id, inputSchema, mcp, permissions, scope` — **no `description`. 0 of 280 had one**, and only 16 property-level descriptions existed across the whole catalog. Generated CLI help therefore read `Invoke the views.convert capability`.
- **Hand-written MCP tool definitions** — ~12,389 lines across `apps/api/src/mcp/*-tools.ts`, each hand-authoring name, description, a full input schema with per-property prose, annotations, access, anonymized and scope.

The hand-written layer largely existed *because* the capability layer had nowhere to put a description. That prose is prompt engineering, so it was **moved verbatim**, never regenerated.

Since the primary consumer of the CLI is expected to be an AI agent, this prose serves CLI `--help`, MCP listings and chat equally — it belongs on the handler, not in one surface's private layer.

## Before / after

`stella audit-logs list --help`:

```diff
- Invoke the audit-logs.list capability
-      [--workspace-id]   (optional) string
+ Read the organization's audit trail (compliance view). Returns audit entries newest
+ first, each with its action, resource type and id, actor user id, workspace,
+ timestamp, and change detail. …
+      [--workspace-id]   Only entries scoped to this matter/workspace [(optional) string]
```

## Coverage

- Tool-level: **21 / 280** (was 0)
- Property-level: **105** (was 16)
- Ratchet baseline seeded at the real number, **518** across both mirrors, may only decrease

The 21 is bounded structurally, not by effort: only 33 capabilities have a 1:1 curated twin, 46 are `covered`, and **201 have no hand-written twin at all** — there is no prose in the repo to move for them. Reaching 280 means *authoring* descriptions, which is a separate task.

## Design note

Per-property descriptions needed **no new field**: an Elysia `t.X({ description })` already reached the catalog's `inputSchema` and was simply never consumed. The fix was in `classifyProp`. That also gave the 44 curated-tool commands their existing property prose for free.

## Prose deliberately NOT migrated

- **12 tool-disposition capabilities skipped** because the tool spans more than its backing handler, so moving the string verbatim would make it *wrong* — typically a create-or-update tool behind a create-only handler. `delete_document` is a harder mismatch (one document, optionally one version, vs a handler taking `entityIds: SafeId[]`).
- **Chat/MCP-projection-only prose dropped, not silently**: agent-sequencing guidance ("Use this after matter overview surfaces a contact…"), cross-tool references by name, and `confirm` properties that are an MCP-only gate with no handler field.
- `set_practice_jurisdictions` keeps agent-instructional phrasing that reads oddly in CLI help — moved verbatim per move-don't-rewrite, flagged for a wording pass.
- Cursor descriptions are inconsistent (some generalized, some skipped); worth one normalizing pass.

## Guard

`capabilities-without-description` ratchet, counting both mirrors, whitespace-only treated as missing so a blank string cannot silence it. Proven to fire by deleting a description and re-exporting:

```
capabilities-without-description: 518 -> 520 (+2)
→ exit 1
```

Restored: `--check: OK`, `--self-test: PASS`.

## Verification

Regression baseline — `apps/api` suite on a throwaway `origin/main` worktree vs this branch, same runner:

| | pass | fail | skip | files |
|---|---|---|---|---|
| `origin/main` | 4641 | 0 | 4 | 491 |
| this branch | 4641 | 0 | 4 | 491 |

**Net regressions: zero.** Plus `packages/cli` 355 pass, `apps/api/scripts` 165 pass, both typechecks clean, mirrors byte-identical, exporter `--check` OK.

**Not verified:** the chat/ref invariant is preserved by construction (no file under `handlers/chat/` or the ref registry appears in the diff, and descriptions are static prose that never touches ref mediation), but no live chat turn was exercised.

## What this does and does not achieve

Honest framing: this collapses the **description** layer for 21 tools, not the tool layer.

Of the 12,389 lines in the tool files, **8,489 are definition blocks and ~3,900 are real handler logic** — MCP arg mapping, ref mediation, anonymization field specs. That is not duplication and cannot be deleted. Within the definitions, only the description strings for those 21 are now redundant, and only once the MCP listing path reads from the catalog. `annotations`, `access`, `anonymized` (redaction policy, MCP-only), `scope` and the snake_case↔camelCase mapping have no catalog equivalent.

So "one registry" is now true for descriptions; it is not yet true for tool definitions, and deletion alone will not make it so.


jan-kubica:
Applied — and it is a forward-looking trap worth spelling out.

`TokenHandlerConfig` and `PublicHandlerConfig` did not extend `CapabilityDescription`, so those factories could not declare one. No token/public handler is catalogued today (0 of 280; all are `mcp: internal`), but the exporter *does* discover both kinds — `FACTORY_KIND_PATTERNS` has entries for them. So the first token or public handler exposed as a capability would be counted by the new `capabilities-without-description` ratchet **with no way to satisfy it**: the guard would fire and the fix would require a type change.

Both configs now extend it. Their route contexts also strip `description` alongside `mcp`, matching `ConfigRouteSchema` on the workspace path, so it is never mistaken for a route schema slot — that asymmetry would have been the next surprise.

Verified: `apps/api` typecheck clean. `src/lib` + `src/mcp` suites run with and without the change (stash baseline) — identical 10 pre-existing failures, all network/upstream-dependent. **Net regressions: zero.**

PR #1293, opened 2026-07-20
feat(api): machine API keys for non-interactive CLI and MCP auth

jan-kubica:
## Problem

There was no way for a machine to authenticate. `stella auth login` is browser-only PKCE loopback, and the CLI read no credential env vars, so CI, containers and headless agents could not authenticate at all.

An earlier attempt targeted the `client_credentials` grant and had to be abandoned: `@better-auth/oauth-provider` builds the access token as `{ ...customClaims, sub: user?.id }`, and `handleClientCredentialsGrant` passes no user — so `sub` is dropped entirely and `org_id` comes back undefined. `apps/api/src/mcp/auth.ts` hard-requires both. Making it work would have meant either weakening MCP verification or patching the auth dependency.

`@better-auth/api-key` is the purpose-built plugin for this and needs neither.

## What this adds

- `apikey` storage, plugin registered with its own HTTP CRUD disabled (`/api-key/create|update|delete` in `disabledPaths`) so lifecycle only flows through first-party handlers that enforce org rules.
- Org-scoped lifecycle endpoints: create, list, rotate, revoke — gated on `organizationSettings: ["update"]` with the org taken from `session.activeOrganizationId`, never the body.
- MCP accepts a key as a credential type, resolving to the same `{userId, organizationId}` the JWT path produces, so it lands on the identical member and RLS authorization. No existing check was relaxed.
- CLI authenticates non-interactively via `STELLA_API_KEY`, with documented precedence over a stored credential.
- Keys purged on account deletion; RLS matrix entry added.

## The gap this PR closes

An earlier revision was caller-scoped, not org-scoped: `loadOrganizationMachineApiKey` called the plugin's `getApiKey` with the **caller's headers**, so the plugin's `referenceId === session.user.id` check applied even though the handler was org-permission-gated. Net effect: **an org admin could not revoke a key minted by a different admin** — a credential that cannot be revoked when its creator leaves, or when it leaks.

Reads now query the `apikey` table directly with the organization filter **in SQL**, never post-filtered in JS:

```ts
const organizationScope = (organizationId: SafeId<"organization">) =>
  and(
    eq(apikey.configId, MACHINE_API_KEY_CONFIG_ID),
    sql`${metadataOrganizationId} = ${organizationId}`,
  );
```

Authorization is unchanged; only the read widens. Revoke passes the key's real `ownerUserId` to the plugin's server-side path with no headers, since headers would re-impose caller scoping. Foreign-org keys 404 rather than 403.

`list` returns the standard `Page<T>` envelope via `createCursorPage`, keyset on `(created_at, id)`, paginating before dropping undescribable rows so `nextCursor` stays anchored to a real row.

## Verification

The new tests were confirmed **failing before the fix** — run against the pre-fix commit in a throwaway worktree: 5 fail / 3 pass, with the failure output showing the old JS post-filter `metadata.output.organizationId !== session.activeOrganizationId`. After: 8 pass / 0 fail.

| Check | Result |
|---|---|
| `apps/api` typecheck | exit 0 |
| `apps/api` suite | 4664 pass / 0 fail / 4 skip |
| `origin/main` baseline | 4641 pass / 0 fail / 4 skip |
| **Net regressions** | **zero** (+23 tests) |
| `packages/cli` | 359 pass / 0 fail |

The migration's two expression indexes were applied to a throwaway Postgres: both create successfully, and `EXPLAIN` confirms `Bitmap Index Scan on apikey_org_keyset_idx` with the org predicate pushed into the index.

## Not verified

- No clean **full**-suite run after the final refactor. Later runs hit DB-hook timeouts in `hosted-usage-webhook/dispatch.test.ts` and `operator/query.test.ts` plus a timing-sensitive `runSandbox` test; these reproduce on pristine `origin/main` under the same load (load average 18), so they are load-induced flakes rather than regressions — but the clean numbers above were taken before the machine degraded.
- Migration not applied to the shared dev DB (throwaway probe only).
- The `Page<T>` envelope is covered by typecheck and unit tests, not exercised end-to-end over HTTP.
- No web typecheck, no `bun run verify`, no browser checks.

## Summary by CodeRabbit

* **New Features**
  * Added machine API key management, including creation, listing, rotation and revocation.
  * Machine API keys now support organisation-scoped authentication for MCP requests.
  * Added CLI support for `STELLA_API_KEY` as the preferred credential (including updated `whoami` output).
* **Security**
  * Enforced permission validation, audit logging, and cross-organisation isolation for machine API keys.
  * Machine API keys are disabled during member/account deletion and are protected by control-plane least-privilege.
* **Bug Fixes**
  * CLI identity reporting now clearly indicates when stored credentials are bypassed by `STELLA_API_KEY`.


jan-kubica:
Both findings are already addressed on the current head:

**`CONCURRENTLY`** — the review is against an older commit. Both index creations are `CREATE INDEX CONCURRENTLY`, with the `COMMIT`/`BEGIN` split Drizzle needs, `SET statement_timeout = 0` / `SET lock_timeout = 0` around the concurrent section, and the `squawk-ignore transaction-nesting` comments — matching `20260707120000_property_role`. Squawk's `check` job passes on that commit.

They are also drop-then-create rather than `IF NOT EXISTS`: a cancelled concurrent build leaves an INVALID index, which `IF NOT EXISTS` would skip recreating, leaving the tenant filter on an index Postgres will not use while the migration records as applied.

**`defaultNow()`** — correct, and this was the actual `apply-fresh` failure. The migration creates `created_at`/`updated_at` as `DEFAULT now() NOT NULL` while the schema declared only `.notNull()`. Fixed, and verified against a throwaway Postgres: `drizzle-kit push --explain` reports `No changes detected`.

Worth noting the expression indexes were **not** the mismatch — they round-trip through drizzle with zero diff. The timestamp defaults were.

jan-kubica:
Fixed — thank you, this was a real gap and the re-invite case is the sharp end of it.

`revokeOrganizationMemberAuthArtifacts` now disables the member's machine keys in the same transaction as the OAuth token revocation, scoped by:

```sql
reference_id = :userId
AND config_id = 'machine'
AND (metadata::jsonb ->> 'organizationId') = :organizationId
AND enabled = true
```

Both ownership halves are load-bearing for the multi-org case: owner-only would revoke keys in organizations the user has **not** left, org-only would revoke their colleagues' keys. The predicate now lives in one shared module used by both the read queries and this revocation, so the two cannot drift apart.

Disabled rather than deleted, consistent with the `revoke` handler: the audit trail and the `start` prefix are what let an operator match a leaked credential back to its key. `resolveMachineApiKeySession` checks `enabled` *before* the membership lookup, so a disabled key stays dead across a re-invite — which is the specific hole you identified.

Tests were confirmed failing before the change (5 pass / 2 fail → 7 pass / 0 fail), the two failures being exactly "removing a member revokes the machine keys they hold in that organization" and "re-inviting the member does not resurrect their old keys". They run against real SQL on PGlite rather than a mocked transaction, which for a tenant predicate is the only test worth having. Three further tests guard against an over-broad fix (keys in another org, a colleague's keys, another config) and pass in both directions.

Suite: 4673 pass / 0 fail against a 4641 baseline on pristine main. Typecheck, oxlint and oxfmt clean.

jan-kubica:
Worked through the codex findings — verified each against the current head rather than applying blind. Three were live and are now fixed; three were already resolved or don't apply.

**Fixed:**
- **Rotate of a revoked key** (P2) — a rotate racing or retried after a revoke minted a fresh active key carrying the revoked key's scopes, undoing the revocation. `rotate` now rejects a disabled key (409) before minting. Pinned structurally: a test asserts the `enabled` guard precedes the mint call.
- **Prototype-key permission resources** (P2) — a resource named `constructor` / `__proto__` resolved to a prototype value under `STATEMENT_ACTIONS[resource]`, so `.includes` threw a 500 instead of returning unknown-resource. Guarded with `Object.hasOwn`. Test confirmed failing before the fix.
- **Key-name length mismatch** (P2) — the boundary schema accepted 64 chars while the plugin defaults `maximumNameLength` to 32, so a 33-64 char name passed our validation then hit the plugin's own 400. One shared `MACHINE_API_KEY_NAME_MAX_LENGTH` constant now drives both the schema and the plugin config.

**Not applicable / already done:**
- **Disable `/api-key/verify`** (P2) — this plugin version exposes no such HTTP route. Its `createAuthEndpoint` registrations are exactly `create/update/delete/get/list`, all already in `disabledPaths`; `verifyApiKey` is a server-side function only. No oracle to close.
- **Revoke on member removal** (P1) — already fixed in `4bf5b8f4fe` (the shared-predicate cleanup in `revokeOrganizationMemberAuthArtifacts`).
- **Reset timeouts / avoid `IF NOT EXISTS` on the concurrent build** (P2) — already addressed: the migration sets `statement_timeout = 0` for the concurrent section and uses drop-then-create rather than `IF NOT EXISTS`.

apps/api typecheck, lint, and format all clean; the two affected security suites pass 27/27.

PR #1295, opened 2026-07-20
fix(api): represent text file attachments as text and gate document parts by model modality

jan-kubica:
## Problem

Chat file attachments are hydrated into AI content parts before being sent to the model. Two issues in that path:

1. **Text-extractable formats were sent as `document` parts** (or, in the default non-anonymized mode, as raw bytes). `createChatAttachmentPart` types any non-image mime as `type: "document"`, so docx/txt/csv/md all became `document` parts — and a `document` part carrying non-PDF bytes is not valid input for several provider adapters. Depending on the selected model, such an attachment could fail the chat stream.
2. **No modality gate.** Chat model selection never checked whether the resolved model can actually ingest an attachment's format, so an incompatible pairing reached the adapter and threw.

The raw-send path made it worse: the `rawOverride` branch returned before the extraction code, so whenever anonymization was off (the default), a docx was shipped as raw bytes that no adapter reads.

## Change

- **Text-extractable attachments (docx/txt/csv/md) are always reduced to a `text` content part**, in both send modes, using folio's structure-preserving Markdown extraction (`extractMarkdown` — headings, tables, lists) instead of a flat paragraph join. Text is universal across provider adapters, so these attachments are never modality-gated and cannot crash a stream. The separately-persisted `userfile://` reference part still renders the attachment chip in the UI; only the provider-bound copy changes.
- **Genuine `document` (PDF) attachments are gated by the resolved model's input modality** before dispatch: a model that cannot ingest the format gets a clean `422`, and a fallback model that would also reject it is dropped so a failover cannot re-introduce the mismatch.
- **Mistral vision models now accept PDF `document` input** (adopts [TanStack/ai#912](https://github.com/TanStack/ai/pull/912) as a local patch until release). This is exposed through a chat-attachment capability set (`CHAT_PDF_ATTACHMENT_MODEL_OPTIONS`) kept **distinct** from the pdf-role catalog, so it does not turn Mistral into a pdf-role provider elsewhere in the app.

## Why it's structurally safe going forward

- Text-extractable content is now *text*, so it can never be a modality mismatch — the only `document` parts reaching a model are PDFs, which are gated.
- A **catalog-wide property test** pins every offered model's document capability (textual + PDF) to the typed catalog, with a superset invariant, so adding a model/provider without wiring its modality fails the test instead of shipping a latent crash.
- Extraction tests assert text-extractable attachments become `text` parts (never raw/`document`) in both modes; gate tests cover the PDF/textual/refused matrix.

## Tests
`@stll/ai-catalog` and the affected `@stll/api` suites pass (property, gate, extraction, and the existing hydration/boundary tests).

## Summary by CodeRabbit

- **New Features**
  - Added gated support for PDF and text document attachments on compatible chat models, including selected Mistral models.
  - DOCX uploads are consistently reduced to extracted, readable text before being sent.

- **Bug Fixes**
  - Unsupported attachment MIME types are rejected before chat generation begins (HTTP 422).
  - Improved failover so fallback models don’t retry with attachments they can’t handle.
  - Enhanced extracted-text handling, including preservation of filenames and correct empty-content behaviour.

- **Tests**
  - Added/updated coverage for document-attachment capability gating and DOCX extraction.

### DIFF
--- a/apps/api/src/handlers/chat/attachment-modality.test.ts
+++ b/apps/api/src/handlers/chat/attachment-modality.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  TEXT_CSV_MIME_TYPE,
+  TEXT_MARKDOWN_MIME_TYPE,
+  TEXT_PLAIN_MIME_TYPE,
+} from "@/api/handlers/chat/attachment-validation";
+import { DOCX_MIME_TYPE, PDF_MIME_TYPE } from "@/api/mime-types";
+
+const { modelAcceptsDocumentAttachment } =
+  await import("@/api/handlers/chat/attachment-modality");
+
+// A vision Mistral model: PDF-capable (via document_url) but no textual
+// document support. This is the case that crashed the chat stream.
+const MISTRAL_VISION = {
+  provider: "mistral",
+  modelId: "mistral-medium-latest",
+} as const;
+// A document-capable Anthropic model: accepts both PDF and textual documents.
+const ANTHROPIC_DOC = {
+  provider: "anthropic",
+  modelId: "claude-sonnet-4-6",
+} as const;
+
+describe("modelAcceptsDocumentAttachment", () => {
+  test("allows a PDF to a Mistral vision model but refuses a textual document", () => {
+    expect(
+      modelAcceptsDocumentAttachment({
+        model: MISTRAL_VISION,
+        mimeType: PDF_MIME_TYPE,
+      }),
+    ).toBe(true);
+    for (const mimeType of [
+      TEXT_PLAIN_MIME_TYPE,
+      TEXT_CSV_MIME_TYPE,
+      TEXT_MARKDOWN_MIME_TYPE,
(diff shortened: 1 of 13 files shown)


jan-kubica:
Review follow-up — addressed in the current revision:
- Direct-text attachments now capped to `LIMITS.chatContextFileMaxChars` (like the DOCX branch).
- Modality gate returns a plain 422 instead of a third-party-boundary refusal payload (avoids the "send without anonymization" retry for a model mismatch).
- Removed the stray `.bun-tag` artifact from the patch.
- Block-level content controls (`w:sdt > w:sdtContent`) are preserved in DOCX extraction.

Deferred to the immediate follow-up: the in-repo DOCX→markdown walker is being replaced wholesale by folio-core's DOCX parser, which natively handles every content-control position (block, table-cell, footnote, inline). The table-cell content-control edge lands with that migration rather than expanding a walker that is about to be deleted. Per-model document capability is already pinned exhaustively (all models × mimes) by the catalog-wide property test in `tanstack-ai-models.test.ts`.

PR #1297, opened 2026-07-21
fix(api): resolve capability scope from access, so read-only credentials can read

jan-kubica:
## Problem 
Capability `scope` was resolved **per-domain** — one OAuth scope per domain regardless of read vs write — so every READ capability inherited its domain's WRITE scope. Measured on the committed catalog: **79 of 118 read capabilities required a `*_write` scope**. A credential holding `stella:read` — the natural choice for a safe read-only agent — could invoke almost nothing.

Verified live before and after: a `stella:read`-only machine key was rejected from `stella contacts list` ("requires scope: stella:matters_write"); after this change the same command returns a normal `Page<T>`.

## Fix

Scope now resolves from **(access, domain)**, not domain alone. A read capability resolves to its domain's read scope; a write to the write scope:

| Domain write/consent scope | Read scope |
|---|---|
| `stella:matters_write` | `stella:read` |
| `stella:knowledge_write` | `stella:read` |
| `stella:billing_write` | `stella:read` |
| `stella:documents_write` | `stella:read` |
| `stella:admin_write` | `stella:admin_read` |
| non-tiered (`stella:chat`, `stella:skills`, `stella:templates`, …) | itself (read == write) |

Read caps requiring a `*_write` scope: **79 → 0**.

## The structural guard (makes the class impossible)

1. **Exporter guard**: the export/`--check` aborts if any `access === "read"` capability resolves to one of the 5 write-only scopes, naming the offending id. Because the resolver now keys reads to read scopes, this can only fire if someone breaks the resolver.
2. **Ratchet** `read-capabilities-with-write-scope`, baselined **0**, counting both mirrors, with a `--self-test` fixture — so a regression fails CI even if the exporter guard were bypassed.

Both were proven to fire (break → red with per-id message → restore → green), independently re-verified.

## On write-implies-read (deliberately NOT shipped)

An earlier revision carried an `elevatedByScope` field so a domain write grant would also satisfy that domain's reads. **Live testing showed it was inert**: every capability reaches the runtime scope gate only via `invoke_capability`, whose own scope is `stella:read`, so any session past the transport gate already holds `stella:read` and satisfies a read directly; a write-only key is rejected at the transport before the elevation gate runs. The field carried no behaviour and its test proved a gate the real path never reaches, so both were removed. Real write-implies-read means changing the `invoke_capability` transport scope and both `server-core` gates — a deliberate follow-up if write-only keys ever need read access, not a silently-inert field.

Practical impact of omitting it is near-zero: the default consent bundle includes `stella:read`, so normal agents read fine.

## Verification

- Read caps with a write scope: **79 → 0**; both mirrors byte-identical; no drift; generator-formatted.
- `apps/api` typecheck clean; `bunx oxlint` clean; `oxfmt --check` clean.
- Runtime scope-gate suite 64/0; catalog lib suite 106/0; ratchet `--check` OK, `--self-test` PASS.
- **Live**: `stella:read`-only key lists contacts across domains; a write-only key is (correctly) still denied a read it has no read grant for.

## Summary by CodeRabbit

- **New Features**
  - Improved capability scope handling by distinguishing read and write access levels.
  - Added clearer support for elevated access requirements in capability catalogues.
  - Read capabilities now reject incompatible write-only permissions.

- **Bug Fixes**
  - Corrected scope assignments across multiple capability domains.
  - Improved validation to prevent stale or incorrectly mapped scope overrides.

- **Documentation**
  - Updated capability coverage tables and access descriptions to reflect the revised scope mappings.


jan-kubica:
Both findings addressed — thank you, the codex P1 was a real regression this PR introduced.

**Upload capabilities behind a write scope (codex P1):** correct and serious. `uploads.create/update/delete` mint presigned PUT URLs and finalize file writes, but their route permission is only `workspace:["read"]` because the real write authorization runs inside the handler (`authorizeUploadPurpose`), invisible to verb classification. They classified as `read`, and since this PR makes read resolve to `stella:read`, a read-only consent could have performed file writes.

Pinned them to `access: "write"` via `ACCESS_OVERRIDES` — the same mechanism `flows.run-*` already uses for writes whose route permission looks like a read. They resolve to `stella:matters_write` again (verified in the regenerated catalog), so a read-only token can no longer upload. Added a regression test asserting the upload lifecycle stays `write`/`matters_write`, so dropping the override fails CI.

**`elevatedByScope` CLI leftover (gemini):** right — I removed it from the API exporter and runtime but missed the CLI-side schema/projection/tree type. Stripped all three; no `elevatedByScope` reference remains in `packages/cli/src` or the generated JSON.

Verification: catalog `--check` OK, both mirrors byte-identical, CLI codegen idempotent; `apps/api` typecheck + lint clean; scope-gate + catalog-lib suites 170/0; CLI suite 360/0; ratchet `read-capabilities-with-write-scope` still 0.

jan-kubica:
Addressed the review findings; the PR grew substantially because it now fixes the root cause, not just the symptom.

**views.list / document-types.list seeding (codex P1) — fixed by removing the write, not reclassifying it.** These list handlers seeded default rows on first read (insert + CREATE audit), so `access: read` let a read-only token create views/document-types by listing. Rather than mark them write (which would stop read-only agents from listing), the seeding moved OFF the read path: default views seed at workspace creation, default document types at org creation (`afterCreateOrganization`), with an idempotent additive backfill migration for existing workspaces/orgs. Both list handlers are now pure SELECTs (grep-confirmed mutation-free) and truthfully `read`. Virtual-defaults-on-read was ruled out: navigation relies on persisted view ids.

**The earlier P1s (uploads / organize-suggestions / case-law.analysis.generate reachable with a read scope) are resolved by the core change here.** `access` is no longer inferred from the route permission (which answers *who may call*, a different axis from *reads or writes*). It is now a DECLARED, reviewed field on every handler config. A five-agent security review traced all 97 inferred reads: 84 affirmed read, 13 corrected to write. The export now FAILS on any capability that would reach a read scope without an explicit `access: read` affirmation — strictly stronger than the previous `.authorize()` regex, which missed exactly the write-without-authorize cases you flagged.

**Removed as subsumed:** the ACCESS_OVERRIDES table, its stale-entry guard, and the runtime-authorization regex guard + predicate + test.

Rulings applied (human-reviewed): audit-only and external-memo side effects → read; AI-metered previews → write (`templates.prefill` corrected). Verified: read-caps-with-write-scope = 0, both mirrors byte-identical, typecheck + ratchet + 204 catalog/scope/handler tests green.

jan-kubica:
Both P2s handled:

**seed-dev duplicate views — fixed.** Correct: `workspace_views` has only a primary key, so `onConflictDoNothing()` catches nothing on the fresh ids a reseed generates. Now seeds only workspaces with no existing views, matching the production create path and the backfill migration's NOT EXISTS guard.

**audit-logs.list — staying read, deliberately.** Two reasons. First, its scope is `stella:admin_read`, not `stella:read` — a general read-only credential cannot reach it; only an admin-read token can. Second, unlike the view-seeding case, the ACCESS audit row is not an incidental side effect that can be split off — recording that someone accessed the audit log IS the compliance feature (audit-of-audit); removing it from the read path would defeat the purpose. The row is a truthful record of the access that just occurred, not falsifiable user data, so it is not a mutation hole in the escalation sense. This case was explicitly reviewed and ruled read (it was named in the audit-only category alongside download-zip/table-export). Requiring admin-WRITE consent to READ the audit log would break legitimate read-only compliance monitoring.

PR #1298, opened 2026-07-21
refactor(api): extract chat DOCX to markdown via folio-core, drop the local walker

jan-kubica:
## What

Now that `@stll/folio-core` exposes a server-reachable **`docxToMarkdown(bytes)`**, the hand-rolled OOXML→markdown walker in the chat attachment path is redundant. This delegates to folio's DOCX parser and removes the local implementation.

- **Bump** `@stll/folio-core` 0.12→0.14 and `@stll/docx-core` 0.3→0.5 (folio-core 0.14 requires docx-core `^0.4.0`).
- **Swap** chat extraction (`upload-files.ts`) from the local `extractMarkdown` to folio's `docxToMarkdown`.
- **Delete** the local markdown extractor + its tests from `extract-text.ts` (net −466 lines there). `extractText` / `documentTextForAiFields` keep using folio's `extractDocxText`.

## Why folio's version is better

folio's parser handles **content controls at every position** (block, table-cell, footnote, inline) natively — the local walker had dropped some of these — and folio ships fixtures + a seeded fuzz test + real-corpus snapshots guarding content completeness. One source of truth for DOCX→markdown.

## Test fixtures

`templates.test.ts`'s `makeDocxWithParts` previously wrote header/footer parts by filename without wiring them. `extractDocxText` resolves parts through `document.xml.rels` + `sectPr` references (and ignores unreferenced parts), so the fixtures now wire headers/footers the way real Word documents do. This makes the header/footer extraction + discovery tests exercise the real behavior.

Full affected `@stll/api` suite passes.

## Summary by CodeRabbit

- **Improvements**
  - Improved DOCX text extraction for chat attachments and document content.
  - Preserved formatting metadata such as styles, emphasis, font size and alignment.
  - Improved handling of referenced headers and footers in DOCX files.
- **Maintenance**
  - Updated document-processing components to newer versions for improved compatibility.


jan-kubica:
Review follow-up:

- **codex (docx-core range):** correct and important — the root cause is in folio, not here. `@stll/folio-core@0.14.0` imports `normalizeRevisionId`, which only exists in `@stll/docx-core@0.5.0`, yet folio-core declares `docx-core: ^0.4.0` (which excludes 0.5.0). So folio-core@0.14 is effectively broken on a fresh install that resolves docx-core to 0.4.x. The resolution here forces the version folio-core actually needs; all DOCX paths (extraction, templates, chat) plus e2e pass on it. Fixing folio-core to declare `^0.5.0` and republishing is a fast-follow.
- **gemini (test `.replace` fragility):** the helper assumes the `<w:document ` prefix produced by the local `WRAP` (which always emits that exact form), so it is safe as written; noted for if `WRAP` ever changes.

PR #1299, opened 2026-07-21
feat(api): add read_workspace_document chat tool

jan-kubica:
## Summary
- Adds a server-executed `read_workspace_document` chat tool that reads a workspace document's DOCX content as Markdown (folio's `docxToMarkdown`), so the agent can pull any authorized document into context on demand.
- Input: `{ entityId, versionId? }` (`versionId` optional, defaults to the entity's current version). Resolution follows the same TOCTOU-guarded, workspace-authorized pattern as `compare_versions` (`version-compare-tools.ts`): the version query requires `workspaceId IN toolWorkspaceIds` and `deletedAt IS NULL` in a single read, and the DOCX field lookup happens in the same round trip.
- Output is capped at `LIMITS.chatContextFileMaxChars`.
- Classified `internal` (read-only, no per-call approval) and registered whenever the caller has at least one accessible workspace — not tied to the active file, unlike `compare_versions`.
- Distinct from the folio-agents `read_document` tool (`folio-agent-tools.ts`): that one is client-executed against the *live DOCX editor* for the active file only. This tool is server-executed and can read any authorized document by id.

## Test plan
- [x] `bun --filter @stll/api typecheck`
- [x] `bun test --preload ./src/tests/setup-env.ts src/handlers/chat/tools/read-workspace-document-tools.test.ts` (7 pass)
- [x] `oxlint --deny-warnings --type-aware` on changed files
- [x] `oxfmt` on changed files

### DIFF
--- a/apps/api/src/handlers/chat/tools/chat-tools.ts
+++ b/apps/api/src/handlers/chat/tools/chat-tools.ts
@@ -46,6 +46,10 @@ import {
 import { createInfosoudTools } from "@/api/handlers/chat/tools/infosoud-tools";
 import { createOrgTools } from "@/api/handlers/chat/tools/org-tools";
 import {
+  createReadWorkspaceDocumentTools,
+  READ_WORKSPACE_DOCUMENT_TOOL_NAME,
+} from "@/api/handlers/chat/tools/read-workspace-document-tools";
+import {
   buildChatWriteTools,
   type ChatRegistryWriteToolMap,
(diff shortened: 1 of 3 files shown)


jan-kubica:
Superseded by upgrading read_content_across_matters (the chat agent's existing document-read path) to return folio Markdown directly, converging onto a single read path instead of adding a parallel tool — see #1303.

PR #1335, opened 2026-07-22
fix(cli): unify contact discovery and output flags

jan-kubica:
Promotes the organization contact directory to a curated `list_contacts` MCP tool and the canonical `contact list` CLI command while reusing the HTTP endpoint's tenant-scoped query. The endpoint disposition now identifies the curated tool, so coverage prevents it from drifting back into an orphaned or duplicate generic capability.

Centralizes output flags for generated tools, capabilities, resources, and the hand-authored upload command. Adds registry invariants that require curated detail readers to have a discovery sibling, plus focused MCP and CLI coverage.

CC on behalf of jan-kubica


jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

PR #1349, opened 2026-07-23
fix(web): chat overlay QA polish and composer (+) menu parity

jan-kubica:
Focused QA pass on the Folio DOCX chat overlay + inspector, plus a couple of adjacent UI fixes.

## Chat overlay / review
- **Editing-blocked toast → quiet chip.** The "Editing blocked" toast fired on every edit attempt against a DOCX Folio can't safely rewrite (a real corruption guard). Dropped it on all three edit-attempt paths (browser editor + inspector edit-session + tab) and surface the state quietly on the composer's edit-mode control as a **View only** chip (tooltip carries the reason). The composer placeholder no longer promises "…or edit" when the file can't be edited.
- **Revert now works after later edits.** Review *Revert* used `undoDocumentOperations` — a strict LIFO/exact-document-match undo that rejects (`Action failed`) after any subsequent edit, another accept, or resolving the change through the Word review controls. It now rejects the tracked change by its stored **revision ids** (`rejectAIEditOperation`), which is position-independent (mirrors `playbook-facet.tsx`). Direct-mode accepts still fall back to the stack undo.
- **Failure toasts no longer hidden.** The toast viewport was `z-[60]`, below the floating composer's blur veil (`z-[80]`); raised to `z-[90]`.

## Thread card / composer
- **Follow-up chips** moved into the transcript scroll flow (they were a pinned footer that the card's `overflow-hidden`/`max-h` clipped), so they stay inside the chat window and scroll naturally.
- **Autoscroll rewrite.** The old effect only fired on `[isGenerating, lastMessageId]`, so it did nothing while content streamed and then jumped once at the end — never re-firing when the async follow-up chips loaded, leaving them below the fold. Replaced with a stick-to-bottom that follows content growth (streaming, reasoning expand, chips) while pinned, re-pins on a new message, and unpins when the user scrolls up.
- **Reasoning block** folds by default and is collapsible once the message stops streaming (previously it stayed a non-collapsible view whenever a message had no answer text).
- **Collapse control** moved into its own header row so it no longer collides with the transcript scrollbar.
- **Matter picker** truncates on narrow docks (added `min-w-0`) instead of overflowing the status row.
- **Composer (+) menu parity.** The file-chat `PromptBar` only wired "Attach"; it now forwards the same `models` / `skills` / `context` / `mcp` submenu props to the shared `ComposerPlusMenu` as the main chat, using data already in scope. Other `PromptBar` consumers (inspector chat tab, Template Studio) are unaffected — the props are optional.

## Sidebar / knowledge
- **Recents no longer jump on click.** `selectRecentWorkspaces` forced the active workspace to the top, so selecting a matter yanked it to position 0 on every click. Order now follows activity only; the active matter keeps its activity-ranked slot (surfaced at the bottom only if it would otherwise fall outside the window). Removed the "Recent matters" header.
- **Knowledge catalogue cards** are equal height within a row.

## i18n
- Added `folio.viewOnly` with translations across all locales.

_Also includes an incidental oxfmt pass over some docs/plans/config files._

## Summary by CodeRabbit

* **New Features**
  * Unsafe DOCX files are shown as non-editable **“View only”** via a locked control, with new translated labels.
  * Composer chat model selection is now integrated and send actions are gated while a new model is being persisted.
* **Bug Fixes**
  * Chat transcript “stick to bottom” behaviour is more reliable during streaming.
  * Assistant reasoning now collapses appropriately when no answer is produced.
  * Reverting accepted AI suggestions restores the expected editor changes.
  * Recent items keep activity-based ordering while still surfacing the active item; toast layering is improved.
* **Documentation**
  * Plan/design documents were reformatted for consistency.


jan-kubica:
Addressed the review findings in `01fec07`:

- **Unsafe DOCX still exposed the edit tool** (codex P1): `resolveActiveDocxEditModeState` now returns `unavailable` when the doc is unsafe, so `getActiveFile` reports `supportsDocxEdits: false` and auto mode can't version a file that can't round-trip. The "View only" chip is unchanged. Regression test added.
- **Model-send race** (codex P2 / CodeRabbit Major): both file-chat send paths (composer + follow-up chip) now `await modelSelection.awaitPendingSelection()` and abort on failure, matching the main chat.
- **Revert recovery apply-mode drift** (CodeRabbit Major): the re-apply now uses the mode the change was originally accepted under, threaded via `ApplyOutcome.appliedMode`, instead of the picker's current mode.
- **`selectRecentWorkspaces` limit contract** (CodeRabbit Minor): returns `[]` for a non-positive limit. Regression test added.

The remaining CodeRabbit comments are on `.agents/plans/*` files that this PR only ran through `oxfmt` (formatting-only, no authored changes), so those pre-existing prose nits are out of scope here.

jan-kubica:
Addressed the two follow-ups in \`a3e2a175\` with a single change — modelling DOCX edit-safety as a named \`DocxEditSafety\` (\`"safe" | "checking" | "unsafe"\`) instead of a boolean:

- **codex P2 (pending compatibility still editable):** while the probe is `checking`, the overlay now withholds the AI edit tool (and the mode chip), so a prompt sent in that window can't auto-edit an eventually-unsafe document — matching the native unlock paths.
- **CodeRabbit nit (named state):** the boolean conflated "unchecked" with "safe"; the three-state union keeps them distinct (only `safe` unlocks editing, only `unsafe` shows "View only"). Added `getDocxEditSafety` next to `getDocxEditBlockReason` and a `checking` regression test.

PR #1362, opened 2026-07-24
fix: publish packages only from release metadata

jan-kubica:
## Summary

- trigger npm publishing from Changesets-owned package changelogs instead of package manifest edits
- enforce that automatic publish paths cannot regress to package.json
- satisfy the release contract lint on the TS7/Oxc main baseline

## Validation

- bun install --frozen-lockfile
- targeted release contract tests (5 passed)
- actionlint
- zizmor
- git diff --check

The custom npm publishing workflow and package matrix are otherwise unchanged.

CC on behalf of @jan-kubica

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -18,9 +18,11 @@ on:
   push:
     branches: [main]
     paths:
-      - packages/conditions/package.json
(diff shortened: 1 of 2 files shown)


jan-kubica:
@codex review

CC on behalf of @jan-kubica

PR #1379, opened 2026-07-26
fix(case-law): keep metadata extraction linear in document size

jan-kubica:
Three ingestion regexes backtracked super-linearly against the documents they scan. Because the ingestion daemon is single-threaded, a single oversized document blocks every adapter's cycle, timer and abort signal for as long as the scan runs. Fixing those led to clearing the whole class from the repo, so the second commit is a repo-wide sweep.

## The patterns that started it

**`cz-us`** is the worst. `/(?<judge>\S+(?:\s+\S+){0,2})\s+\(soudce\s+zpravodaj\)/iu` led with the name and ran against the raw page. NALUS pages carry an ASP.NET `__VIEWSTATE` field — one unbroken 12-18 KB run of non-space characters — so every start offset re-walked the rest of that run. Measured on a synthetic blob: 2 k chars 665 ms, 4 k 2.3 s, 8 k 9.5 s, 12 k 18.4 s.

The fix matches the `(soudce zpravodaj)` annotation first, then reads a bounded window back for the name, and takes the decision body rather than the page so the hidden form fields are never scanned. The full RTF/AST parse of the same pages is 9-68 ms, so this was the entire cost.

**`cz-nss`** result-count patterns used `\d[\d\s]*` immediately before `\s+`; the overlapping classes let the engine re-split every trailing space. 1,514 ms on a 16 KB page, ~0 ms after keeping the group separator disjoint.

**`cz-ns`** led with the honorific and lazily expanded `[\p{L}\s,.-]+?` toward a `předsed[ay] senátu` lookahead. A decision signed by a *předsedkyně* matches nothing and pays a full quadratic scan to discover that: 10,270 ms on a 232 KB decision, 0.74 ms after anchoring on the label. Output verified identical across the signature shapes it handles, including stacked titles (`doc. JUDr. …`).

## Clearing the rest

The second commit takes the count from 36 to 0 and widens the metric from the two big apps to every hand-written source file, so the ratchet is a ban rather than a budget. A baseline of 36 cannot distinguish a newly added bomb from an old one in the same file; 0 can.

Every rewrite is equivalent by construction and checked, not eyeballed:

- **21 `\s*[:.]?\s*$` heading tails** across the segmenter and the CZ/SK/PL parsers. Two adjacent `\s*` let the engine re-split trailing whitespace; `\s*(?:[:.]\s*)?$` accepts exactly the same language, verified exhaustively over 4,704 generated strings.
- **Trailing-run strips** (`/\/+$/`, `[.,;:!?]+$`, …): a negative lookbehind pins the match to the first character of the run, so only one start offset can succeed.
- **Scans whose class could span their own delimiter** (`<[^>]*>`, `\{[^}]+\}`, `\[\[[^\]]+?\]\]`): excluding the opening delimiter stops a failed attempt overlapping the next candidate. `stripHtml` output is byte-identical over 18.2 MB of stored court responses.
- **`^\s*` under `m`**, where `\s` crosses newlines and defeats the anchor: `[ \t]*` is both linear and what the code meant.
- Remaining one-offs (ÚS registry sign, NS case-number scan, DOCX directive line) were each diffed against the old pattern over realistic inputs. The DOCX one now matches the marker grammar in `@stll/template-conditions`, which already had the safe form.

## Guards

`extractJudge`'s cost is asserted not to track the size of the hidden form payload, so a regression is a failing test rather than a slow crawl.

The daemon's lag watchdog could not have caught this class. It required `WATCHDOG_MAX_STARVED_TICKS` consecutive starved ticks, but one long synchronous block delivers no tick at all while it runs and then exactly one very late tick, after which the loop is idle and the next tick is punctual — the streak resets and never forms. It now also trips on a single conclusive block, and logs the adapter cycles in flight, so a stall names what held the loop instead of what timed out behind it.

`scripts/ratchet.ts` gains a `super-linear-regexes` metric backed by `scslre`, the analyser behind eslint-plugin-regexp's super-linear rules. Both of its report kinds are counted: the pattern above is a `Move` finding (each retry re-walks the input), not the better-known `Trade`, so gating on `Trade` alone would have missed it.

The counter parses rather than scanning lines, which is a deliberate departure from the rest of that file. It first inherited the shared line-based literal-blanking, which reads the `"` inside `/\s*scale="[^"]*"/u` as opening a string and eats the rest of the pattern — seven regexes were invisible to it for no reason but containing a quote. A guard a new regex evades by quoting itself is not a guard; the self-test fixture now carries quote-bearing patterns to keep that closed.

## Notes

`redos-detector` was evaluated first and rejected: it cannot parse named capture groups, which this codebase uses throughout, and its default scoring flagged 150 of 766 literals including plainly linear ones.

## Verification

- `bun run test` in `apps/api` (3,705 tests) and `apps/web` (1,088 tests)
- `bun scripts/ratchet.ts --check` and `--self-test`; the metric reports 0 across 978 literals repo-wide
- Adapter changes exercised against saved real responses from each court

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved parsing of court decisions, including judge names, case numbers, section headings and document structure across Czech, Slovak, Polish and EU sources.
  - Corrected handling of large result counts, whitespace, punctuation and malformed or nested text.
  - Improved URL, document directive, spreadsheet, stamp and suggestion-text sanitisation.
  - Reduced performance risks when processing large court-document pages.

- **Monitoring**
  - Enhanced ingestion watchdog diagnostics and failure handling, including identification of active processing cycles.

- **Quality**
  - Added automated checks for potentially inefficient regular expressions and expanded regression coverage.


jan-kubica:
Six of the eight findings were valid and are fixed in 531bb6d. Two are not; evidence below.

**Fixed**

- **cz-ns: first-label-only (codex).** Correct, and worse than described — I checked against the committed `cz-ns-page.json`: anchoring on the first `předseda senátu` matched the opening `složeném z předsedy senátu JUDr. …` instead of the signature block, losing the judge on **7 of 2,211** decisions. Now walks every label. One further case needed trailing-whitespace trimming before windowing: a signature sits ~200 spaces above its label and fell outside the 200-char lookback.
- **cz-ns: feminine label (CodeRabbit).** Added `kyně`. Combined with the above, the fixture now shows **0 lost, 6 gained** — the `předsedkyně`-signed decisions never had a judge extracted, before this PR or after my first attempt.
- **Watchdog attribution race (CodeRabbit).** Correct and important. A synchronous block ends before any timer runs, so the cycle's cleanup (a microtask) clears its key before the watchdog callback (a macrotask) reads the set — it could report `""` for the very stall it exists to attribute. A second set, drained by the watchdog rather than by the cycle, keeps the culprit visible.
- **`matter-reference` braces (codex + CodeRabbit).** Correct: `{{SEQ}}` and `{SEQ}{BAD{}` validated as one recognized token. Leftover braces are now rejected explicitly, with cases covering both.
- **Ratchet hardcodes TSX (codex).** Correct, and precisely the failure mode this guard is supposed to be immune to: a misparsed generic arrow in a `.ts` file can swallow the rest of the file and hide a regex. Dialect now comes from the filename, which the counter signature already supplied.
- **Globs stop at `src` (codex).** Correct — the "repo-wide" claim was false. Widening to `apps/*/scripts`, `packages/*/scripts` and e2e surfaced **three more** super-linear patterns (the test runner's type-import strip, two in the capability catalog). Fixed; the metric is still 0 and the capability metrics are unmoved.

**Not valid**

- **Division vs regex literals in the ratchet counter (codex).** Outdated — it cites `e0f6aaa`, the first commit. The lexical `REGEX_LITERAL` it describes was removed in `6315c00`, which replaced it with the TypeScript parser precisely so expression context is decided by a real parser rather than a lookbehind.
- **Inline SQL comments (CodeRabbit).** The behaviour is unchanged. `^\s*--` requires the comment to be preceded only by whitespace from a line start, so the old pattern did not strip inline comments either. Verified directly:

  ```
  input:  'CREATE INDEX x ON y; -- this is not CONCURRENTLY\n-- full line\nSELECT 1;'
  old:    'CREATE INDEX x ON y; -- this is not CONCURRENTLY\n\nSELECT 1;'
  new:    'CREATE INDEX x ON y; -- this is not CONCURRENTLY\n\nSELECT 1;'
  identical: true
  ```

  Stripping inline comments would be a behaviour change, not a fix to this PR.

PR #1380, opened 2026-07-26
fix(case-law): reach the SK Constitutional Court without OAuth2

jan-kubica:
The `sk-us` adapter authenticated against the court's DMS search with an OAuth2 client_credentials application whose id and secret the court published in its own JavaScript bundle, carried here as defaults. That application no longer exists — the token endpoint answers `invalid_client` — so every cycle failed at the token step before issuing a single search.

Replacing the credentials is not the fix, because the endpoint is now open:

- `POST /o/v1/dms/search` unauthenticated returns the full result set, paginates (verified at `start=1500`), and covers 1993 through the current year.
- `GET /docDownload/{documentId}` was already unauthenticated and still is.

So the token flow goes away rather than being re-credentialed.

Sending no `Authorization` header is load-bearing, not incidental: a request carrying a header the endpoint cannot verify is rejected with 401, so an unusable token is strictly worse than none. Consequently a 401/403 now means the court has put the endpoint back behind authentication — which needs a real credential, not a token refresh — and it is surfaced instead of retried.

Also removes the adapter's two `process.env` reads along with the credentials, so it no longer needs its entry in the `forbid-process-env-outside-env-ts` allowlist. Leaving a stale entry there would silently permit future direct env reads in this file.

## Verification

Ran the adapter against the live API for the `2024:0` cursor: 10 decisions returned, each with fulltext (11-16 KB) and a parsed AST (38-51 blocks), ECLI and dates populated, cursor advancing to `2024:10`.

`bun run test` in `apps/api` (case-law and security suites included).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Slovak Constitutional Court case-law retrieval by removing unnecessary token-based authentication.
  * Requests no longer include authorization headers when calling publicly accessible search and PDF endpoints.
  * Updated handling of access failures to surface authentication problems immediately, while preserving retry for intermittent non-auth search issues.

### DIFF
--- a/apps/api/src/handlers/case-law/ingestion/adapters/sk-us.ts
+++ b/apps/api/src/handlers/case-law/ingestion/adapters/sk-us.ts
@@ -1,17 +1,21 @@
 /**
  * Slovak Constitutional Court (Ústavný súd SR) adapter.
  *
- * Fetches decisions from the ustavnysud.sk REST API.
- * The API is a Liferay DXP headless service requiring
- * OAuth2 client_credentials authentication.
+ * Fetches decisions from the ustavnysud.sk REST API, a
+ * Liferay DXP headless service. ~52,000 decisions from 1993
+ * to present.
  *
- * Search: POST /o/v1/dms/search
- * Auth:   POST /o/oauth2/token (client_credentials)
+ * Search: POST /o/v1/dms/search (no auth)
  * PDFs:   GET  /docDownload/{documentId} (no auth)
  *
- * The client_id and client_secret are public credentials
- * embedded in the court's JavaScript bundle for anonymous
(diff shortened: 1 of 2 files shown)


jan-kubica:
Both review comments were the same finding and both were right: the cycle retried every non-abort error once, so a 401/403 issued two identical unauthenticated requests before the pipeline saw it. That contradicted the throw site's own comment.

Fixed in 2c37796: auth statuses (401/403) are re-thrown before the generic flaky-request retry. There is no credential to refresh, so a second identical attempt can only fail the same way.

Re-verified against the live endpoint after the change: cursor `2024:0` returns 10 decisions with fulltext and a parsed AST, cursor advancing to `2024:10`.

PR #1390, opened 2026-07-27
fix(legal-atlas-runner): run the citation-authority recompute on startup

jan-kubica:
Two small corrections in the case-law ingest daemon, both about work that is scheduled but not necessarily reached.

## Recompute on startup

`citationAuthorityLoop` awaits its full interval before the first pass:

```ts
while (true) {
  await Bun.sleep(CITATION_AUTHORITY_INTERVAL_MS);  // six hours
  …recompute…
}
```

That makes the recompute conditional on the process outliving a six-hour sleep. A process with a shorter lifetime performs it zero times no matter how often it is restarted, so the materialized ranking signal is never refreshed. Restarting more often makes it strictly worse, which is the wrong shape for a maintenance task.

Moving the first pass ahead of the sleep removes the dependency; the interval then does what it was meant to do, which is bound how often the recompute repeats rather than whether it happens at all.

## A durable progress record

The daemon's only self-report is `writeHeartbeat()`, which overwrites `/tmp/ingestion.lock` in place. That carries the latest timestamp and no history, so nothing records whether the daemon has been advancing over time.

The health loop already runs on the shortest interval in the file (30s), so it gets a log line carrying uptime, decisions inserted since start, and active cycle count. The insert counter is incremented where each adapter cycle already reports its result, so it costs nothing extra to maintain.

Putting the record on the health loop rather than inside a unit of work is deliberate, and is the same point as the first change: a line emitted by a periodic job only appears if the process outlives that job's period, so a record tied to the shortest loop stays independent of how long any individual piece of work takes.

## Notes

- No behaviour change for adapters, indexing, or the DB slot semaphore.
- `insertedSinceStart` is process-local and resets on restart, which is what makes it readable next to `uptimeSec`.
- Typecheck and `oxfmt` clean; `bun --filter @stll/legal-atlas-runner test` passes.

## Summary by CodeRabbit

* **New Features**
  * Added richer daemon health telemetry, including uptime, ingestion progress, and the number of active processing cycles.
  * Heartbeat updates now provide clearer visibility into ongoing case-law ingestion activity.

### DIFF
--- a/apps/legal-atlas-runner/src/runners/case-law-ingest.ts
+++ b/apps/legal-atlas-runner/src/runners/case-law-ingest.ts
@@ -334,6 +334,58 @@ const writeHeartbeat = () => {
 };
 
 /**
+ * Pages walked by any adapter since this process started.
+ *
+ * Counts `pagesProcessed`, not `inserted`, for the reason `cycleMadeProgress`
+ * documents: a page parked on a persistent corpus-write failure breaks out
+ * before the page counter increments, but reports rows written on every retry.
+ * Counting inserts would make this rise fastest in exactly the stall it exists
+ * to expose.
+ *
+ * It counts pages walked, not distance travelled: the pipeline increments
+ * before its stagnation check, and a caught-up adapter that re-returns its
+ * current cursor (cz-nss yields `${today}:0` once it reaches today) walks one
+ * page per poll without the persisted cursor moving. That is deliberate here.
+ * Being caught up is healthy — `cycleMadeProgress` treats a completed cycle as
+ * progress even with nothing to show for it — so this is an activity counter
+ * to read beside uptime, not a measure of new ground covered. Do not build a
+ * stall alarm on it without first distinguishing the two.
+ */
+let pagesSinceStart = 0;
+
+/**
+ * Record that the daemon is running, and how far it has got.
+ *
+ * The file heartbeat above is overwritten in place, so it carries only the
+ * latest timestamp and no history. This line is the durable counterpart.
+ *
+ * Written through `logInfo` rather than the OTel logger: that pipeline
+ * deliberately configures no external exporter and mirrors only ERROR records
+ * to stderr, so an INFO record there reaches nothing. A liveness signal that
+ * is silently discarded is worse than none, because the absence looks
+ * identical to a dead daemon.
+ *
+ * It belongs to the health loop rather than to any single unit of work: a
+ * line emitted by a periodic job only appears if the process outlives that
+ * job's period, so tying the record to the shortest loop keeps it independent
+ * of how long any individual piece of work takes.
+ *
+ * Uptime comes from `process.uptime()`, not a wall-clock difference: an NTP
+ * correction would make a `Date.now()` delta jump or run backwards, and this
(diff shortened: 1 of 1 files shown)


jan-kubica:
All four findings were valid and are fixed in 284ad88. Two of them were defects that made the change either useless or actively misleading, so thanks — the first in particular.

**Heartbeat went to a sink with no exporter.** Correct, and this made the whole change a no-op. `otel.ts` states "Intentionally no external log exporter", and `logger.ts` mirrors to stderr only for `SeverityNumber.ERROR`, so `logger.info(...)` produced no output at all. Now written through `logInfo`, which is stdout-backed like the rest of the runner's reporting. Noted in the comment as well, because a discarded liveness record is worse than none: its absence is indistinguishable from a dead daemon.

**`inserted` is not durable progress.** Also correct, and the invariant in `cycle-progress.ts` says so explicitly — a page parked on a persistent corpus-write failure reports rows written on every retry while the cursor never moves. Counting inserts would have made the number climb fastest in exactly the stall the counter exists to expose, which is worse than not having it. Switched to `pagesProcessed`, and the comment now cites `cycleMadeProgress` so the next reader sees why it is not the obvious field.

**Increment before the event write.** Done. Moved to immediately after `runIngestionPipeline` returns, ahead of both the halt-reason handling and the event write. Completed pages are durable regardless of how the cycle ends, so a stalled write must not hide them.

**Uptime clock starts after the S3 refreshes.** Done — hoisted above them so `uptimeSec` covers the whole process life.

Typecheck clean, `oxfmt` clean, `bun --filter @stll/legal-atlas-runner test` 8/8.

jan-kubica:
Verified and correct on the facts: `pagesProcessed++` sits above the stagnation check in `pipeline.ts`, and `cz-nss` re-returns `${today}:0` once it reaches today, so a caught-up poll walks a page without the persisted cursor moving.

I have documented the property rather than changed the field, because I think counting it is right here. Being caught up is healthy, not stalled — `cycleMadeProgress` says as much, treating a completed cycle as progress "even with nothing to show for it". A counter that froze whenever the fleet caught up would read as a stall during normal steady state, which inverts the problem.

What the earlier switch from `inserted` did fix is the genuine failure case, and that still holds: a page parked on a persistent corpus-write failure breaks out before the page counter increments, so it no longer climbs while the adapter is stuck.

So the field is an activity counter to read beside uptime, not a measure of new ground covered. The comment now says exactly that, including the cz-nss path, and warns against building a stall alarm on it without first separating the two cases — which is the mistake the wording would otherwise invite. Renamed the doc line from "pages completed" to "pages walked" to match.

Nothing is alarmed on this value today; the accompanying monitoring change keys on the presence of the line, not its contents.

jan-kubica:
Valid, and fixed. `uptimeSec` now comes from `process.uptime()`.

The value exists to be read while diagnosing a restart, which is precisely the moment a wall-clock delta distorted by an NTP correction would mislead — a lifetime that jumps or runs backwards is worse than no lifetime at all.

It also subsumes the earlier point about starting the clock before the S3 refreshes: `process.uptime()` measures from process start rather than from whatever moment this module happens to capture, so the `daemonStartedAt` variable is gone entirely and there is no longer a placement to get wrong.

PR #1416, opened 2026-07-28
fix: exclude private workspaces from releases

jan-kubica:
## Summary

- disable Changesets version and changelog updates for private workspace packages
- add a release-contract regression guard for the repository setting

Changesets documents private-package versioning as enabled by default; setting `privatePackages.version` to `false` keeps application workspaces outside package release plans while preserving normal releases for public packages: https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#privatepackages-object-or-false

## Verification

- `bun run verify`
- `bun run changeset:version` against the current pending changesets with changelog rendering disabled locally: public package versions changed, while private workspace versions and changelogs remained untouched

CC on behalf of sok0

## Summary by CodeRabbit

* **Configuration**
  * Private packages will no longer be automatically versioned during Changesets releases.

* **Tests**
  * Added validation to ensure release configuration continues to exclude private packages from versioning.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,6 +7,9 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
+  "privatePackages": {
+    "version": false
+  },
   "ignore": [
     "@stll/api",
     "@stll/boe",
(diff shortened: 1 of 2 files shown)


jan-kubica:
@codex review

CC on behalf of jan-kubica

PR #1449, opened 2026-07-29
chore(release): 0.6.12

jan-kubica:
Maintenance release. Improvements to ingestion accuracy.


jan-kubica:
Superseded: recut after the in-flight consolidation merges so the release includes it.

PR #1507, opened 2026-07-31
feat(case-law): walk a source oldest-first until it is caught up

jan-kubica:
A source sorted newest-first cannot be caught up by walking offsets forward. It publishes continuously, so each new decision shifts every later offset and items slide past the cursor unseen — the crawl runs indefinitely without converging. Walking oldest-first does converge, because new decisions land at the end, behind the cursor, and the offsets already walked never move. It is the wrong order to *stay* current in, though, so a source needs both.

Adapters can now declare their walks and which follows which:

```ts
traversal: [
  { name: "backfill", buildRequest: (page) => listRequest(page, "ASC"), then: "live" },
  { name: "live", buildRequest: (page) => listRequest(page, "DESC"), then: null },
]
```

**The active walk lives in the cursor** (`backfill:1200` → `live:0`), not in the caller. If the caller chose it per run, a restart, a redeploy, or a second caller would each have to remember where the crawl got to, and a wrong guess either re-walks the whole collection or skips the tail. Persisting it means the phase survives all three and is self-describing; a caller that wants to force a re-walk still can, by setting the cursor.

The handover happens only when a walk has actually reached the end of the collection, so the switch is a fact about the crawl rather than a guess, and it is logged (`case_law.ingestion.traversal_advanced`).

A cursor written before an adapter declared its walks names no walk, and cannot be carried into one — an offset counted newest-first points somewhere unrelated oldest-first. Such a cursor restarts the first walk, which is what catching up needs anyway. Adapters that declare no `traversal` are untouched and keep `buildRequest` and `offset:` cursors.

sk-courts is the first to declare both. It is the source where this matters most: it publishes ~4.67M decisions and we hold a fraction of them, which a forward walk over a newest-first list cannot close.

### DIFF
--- a/apps/api/drizzle/20260731150000_decision_sheet_number/migration.sql
+++ b/apps/api/drizzle/20260731150000_decision_sheet_number/migration.sql
@@ -1,0 +1,16 @@
+SET lock_timeout = '1s';--> statement-breakpoint
+SET statement_timeout = '5s';--> statement-breakpoint
+
+-- Sheet number within the court file (číslo listu), which some sources append
+-- to the docket as `11 C 153/2025-28`. It identifies where a document sits in
+-- the file, not which case it belongs to, and no citation names it, so keeping
+-- it on `case_number` fragments one case into a reference per sheet and
+-- matches nothing.
+--
+-- Values are not backfilled here: rewriting `case_number` on existing rows is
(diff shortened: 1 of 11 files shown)


jan-kubica:
Superseded by a branch cut from main; the original was branched off the sheet-number work and replayed its merged commits.

PR #1508, opened 2026-07-31
feat(case-law): walk a source oldest-first until it is caught up

jan-kubica:
A source sorted newest-first cannot be caught up by walking offsets forward. It publishes continuously, so each new decision shifts every later offset and items slide past the cursor unseen — the crawl runs indefinitely without converging. Walking oldest-first does converge, because new decisions land at the end, behind the cursor, and the offsets already walked never move. It is the wrong order to *stay* current in, though, so a source needs both.

Adapters can now declare their walks and which follows which:

```ts
traversal: [
  { name: "backfill", buildRequest: (page) => listRequest(page, "ASC"), followedBy: "live" },
  { name: "live", buildRequest: (page) => listRequest(page, "DESC"), followedBy: null },
]
```

**The active walk lives in the cursor** (`backfill:1200` → `live:0`), not in the caller. If the caller chose it per run, a restart, a redeploy, or a second caller would each have to remember where the crawl got to, and a wrong guess either re-walks the whole collection or skips the tail. Persisting it means the phase survives all three and is self-describing; a caller that wants to force a re-walk still can, by setting the cursor.

The handover happens only when a walk has actually reached the end of the collection, so the switch is a fact about the crawl rather than a guess, and it is logged (`case_law.ingestion.traversal_advanced`).

A cursor written before an adapter declared its walks names no walk, and cannot be carried into one — an offset counted newest-first points somewhere unrelated oldest-first. Such a cursor restarts the first walk, which is what catching up needs anyway. Adapters that declare no `traversal` are untouched and keep `buildRequest` and `offset:` cursors.

sk-courts is the first to declare both: it is the source where the gap between what is published and what we hold is widest, and a forward walk over a newest-first list cannot close it.

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple pagination traversal modes with persistent progress tracking.
  * Added bounded live updates, enabling efficient retrieval within a defined recent-results window.
  * Improved transitions between historical backfill and live-result retrieval.

* **Bug Fixes**
  * Improved handling of legacy, invalid, timed-out and failed-page cursors.
  * Ensured pagination progress is preserved when switching traversal modes or reaching a window boundary.

### DIFF
--- a/apps/api/src/handlers/case-law/ingestion/adapters/pagination.test.ts
+++ b/apps/api/src/handlers/case-law/ingestion/adapters/pagination.test.ts
@@ -9,7 +9,9 @@ import { asTestRaw, readTestJson } from "@/api/tests/helpers/test-tool-set";
 import {
   createPagePaginatedFetch,
   decodeOffsetCursor,
+  decodeTraversalCursor,
   encodeOffsetCursor,
+  encodeTraversalCursor,
 } from "./pagination";
 import { mockFetchWithFixtures, saveFixture } from "./test-utils";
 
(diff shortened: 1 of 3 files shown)


jan-kubica:
Both P1s were right and are fixed.

**Skip-ahead lost the walk.** The timeout and 5xx branches wrote a bare `offset:N`, which names no walk, so the next call read it as a restart — one transient failure would have discarded the entire backfill, or dropped `live` back into a full historical crawl. Every cursor this call writes now goes through the mode-aware encoder, and a test pins the reading half of that contract.

**The live walk was unbounded.** It advanced deeper each cycle and would have ended parked at the oldest tail, seeing nothing new — the same non-convergence the ordered walks exist to avoid, reached from the other end. Walks can now declare `windowItems`, the point at which they turn back to their own start; sk-courts' `live` sets it to 5000, comfortably more than the source publishes between cycles, so a slow or skipped cycle still cannot let a decision slip past. Catching up declares no window, since it is meant to cross the collection once.

PR #1510, opened 2026-07-31
fix: restore main, and make the storage module mock complete by construction

jan-kubica:
Two failures on `main`, both independent of any feature work.

**`copy-to-workspace.test.ts` took the api suite down.** It replaces `@/api/lib/s3` wholesale with `mock.module`, so every export anything in its import graph reaches has to be present in the factory. `deleteS3ObjectWithSignal` became reachable and resolved to nothing, which surfaces as `SyntaxError: Export named 'deleteS3ObjectWithSignal' not found` and fails the file rather than the assertion. Added to the factory, with a note about why a wholesale module mock carries that obligation.

**`invalidateCreatedDocumentQueries` failed web lint.** The `map` callback returned a promise without being `async`, which `require-await`-style rules reject.

Verified locally: `copy-to-workspace.test.ts` is 11 pass / 0 fail with the mock fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when newly created chat documents are refreshed, ensuring related data is fully updated before the operation completes.
  - Improved handling of file cleanup during workspace and document operations, helping prevent stale stored files from remaining after changes.

- **Tests**
  - Updated automated coverage to reflect current file-management behaviour and improve the reliability of document copying, duplication and upload workflows.

### DIFF
--- a/apps/api/src/handlers/chat/send-message-side-effects.test.ts
+++ b/apps/api/src/handlers/chat/send-message-side-effects.test.ts
@@ -7,16 +7,17 @@ import type { ChatThreadState } from "@/api/handlers/chat/send-message-thread";
 import type { UploadedChatFile } from "@/api/handlers/chat/upload-files";
 import { AUDIT_RESOURCE_TYPE } from "@/api/lib/audit-log";
 import { toSafeId } from "@/api/lib/branded-types";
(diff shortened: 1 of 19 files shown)


jan-kubica:
superseded

PR #1514, opened 2026-07-31
chore: reduce slice-boundary imports

jan-kubica:
Relocates reusable API helpers from handler slices into domain-owned `lib` modules, and promotes shared web queries, stores, and components out of route-private directories.

This lowers the committed boundary ratchets:

- API cross-handler imports: 121 to 15
- API lib-to-handler imports: 48 to 15
- Web cross-route-private imports: 182 to 44

The change preserves existing handler authorization, tenant scoping, and runtime behavior. It also updates the legal-atlas corpus runner to use the promoted legal-search modules.


jan-kubica:
Addressed the automated review finding in b5feac51bf: the API Docker build, CI image compile smoke, and Knip entrypoint now reference `apps/api/src/lib/files/pdf-worker.ts`. The subsequent `api-image-deps` check passed.

CC on behalf of jan-kubica

PR #1548, opened 2026-08-01
fix: harden provider structured output schemas

jan-kubica:
## Summary

- distinguish tool schemas from structured-output schemas at the type boundary
- retain array-bound guarantees in local validation while projecting provider-safe Google structured-output schemas
- exercise nested structured output across every supported provider and default model role each night
- rotate curated non-default models weekly and fail CI when the scheduled matrix drifts from the exhaustive provider catalog

Validated with bun run verify, the security audit, and the live Google daily capability canary.


jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

PR #1590, opened 2026-08-02
feat(case-law): benchmark every source against its own published total

jan-kubica:
Every adapter now either reports the source's own total for the daily coverage benchmark, or documents exactly why it cannot. Verified live against each source:

| adapter | total | how |
|---|---|---|
| cz-regional | **595,310** (exact) | `/api/opendata/{year}` returns exact per-month counts; summed 2020→current. The old stub's comment claimed no such endpoint exists — it does. |
| cz-us | **103,821** | NALUS reports totals only on a search result page; the session flow (ViewState fields + cookies, one all-inclusive date criterion, redirected results page) reads "z celkem N". |
| eu-ecj | **557,177** | SPARQL count of distinct (work, language) pairs under the same type and XHTML-manifestation filters the page query uses — the exact universe the crawl can ingest, not Cellar's broader case-law class. |
| at-courts | **138,445** | The search envelope carries `Hits` on every page; one minimal request with the crawl's own query. |
| cz-nss | fixed | Its empty-search POST silently no-ops on the source's side; an all-inclusive date range supplies the one criterion the form requires. The portal aggregates several courts and cannot be scoped without executing its scripts, so the count covers the portal's collection — which is also what the crawl walks. |
| cz-ns | unchanged | Already correct: the view count (164,227) matches the view the crawl pages through. |
| sk-us | blind spot documented | The portal widget's search endpoint answers scripted requests with an empty 204 even under browser-identical replay; the doc comment records it instead of dead code. |

Coverage this reveals immediately: cz-regional 92.7%, cz-ns 95.3%, cz-us **69.3%** (~32k gap at the constitutional court), eu-ecj **2.6%** of ingestable work-language pairs (the planned historical re-crawl).

### DIFF
--- a/apps/api/src/handlers/case-law/ingestion/adapters/at-courts.ts
+++ b/apps/api/src/handlers/case-law/ingestion/adapters/at-courts.ts
@@ -435,7 +435,8 @@ export const atCourtsAdapter: SourceAdapter = {
         return null;
       }
       const raw = json.OgdSearchResult?.OgdDocumentResults?.Hits?.["#text"];
-      const parsed = raw === undefined ? Number.NaN : Number.parseInt(raw, 10);
+      const parsed =
+        typeof raw === "string" ? Number.parseInt(raw, 10) : Number.NaN;
       return Number.isNaN(parsed) || parsed <= 0 ? null : parsed;
     } catch {
       return null;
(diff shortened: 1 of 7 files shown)


jan-kubica:
Both findings addressed: the health probe's budget now exceeds the slowest adapter's own timeout (the full-range count is slow on the source's side by design), and both search upper bounds derive from the clock instead of freezing 2030. The nullable hit count in the Austrian envelope is guarded as well.

PR #1627, opened 2026-08-04
chore: drop expired brace-expansion quarantine exclude

jan-kubica:
The temporary quarantine exclusion for `brace-expansion` reached its annotated expiry (2026-08-04T10:00Z); the release-age gate admits the package on its own now, and the quarantine guard fails every run until the stale entry is removed. Guard passes locally after removal.


jan-kubica:
Superseded: the expired exclusion was already removed on main in #1626.

PR #1654, opened 2026-08-04
fix(release): retry the web image attestation before failing the release

jan-kubica:
On v0.6.22 the registry refused the first attestation blob upload with a 403,
after the web image had already been published. The release halted at its last
step, leaving a published image with no attestation and nothing downstream able
to promote it. Re-running the job attested the identical digest 13 minutes
later with nothing changed, so the refusal did not reflect the image, the
permissions, or the package.

The step now retries twice, at 30s and 90s, before failing. The API images
already retry once at 10s via `.github/actions/provenance`; that interval would
not have covered the one refusal observed here, hence the longer backoff.

A third refusal still fails the release. A published image without an
attestation is not a releasable artifact, so this widens the tolerance for a
transient registry refusal without weakening what the step guarantees.


jan-kubica:
@coderabbitai review

PR #1661, opened 2026-08-04
fix(ci): stop a workspace neighbour hiding a first-party package

jan-kubica:
The lockfile scan read a fixed 400 characters past each package key to decide
whether that package resolved from the registry or from the workspace. The
window spills into the entries that follow, so a workspace member on a nearby
line marked its registry-resolved neighbour as a workspace member too, and the
coverage check silently stopped requiring it in the excludes.

It hid 4 of this repo's 14 registry-backed first-party packages. The hidden set
included platform binaries, which is precisely what the guard exists to catch:
a quarantined `optionalDependency` is skipped without a word, so the package
resolves, imports cleanly, and throws on first use.

All 14 are excluded today, so nothing is broken. Nothing was enforcing it
either, which is the part worth fixing.

The scan now reads to the end of the entry's own line. Found while checking
whether this guard was worth sharing with the other `@stll` repositories: their
lockfiles put a workspace member directly after a registry one, which is what
exposed the boundary.

### DIFF
--- a/scripts/check-stll-quarantine-excludes.test.ts
+++ b/scripts/check-stll-quarantine-excludes.test.ts
@@ -56,6 +56,25 @@ describe("quarantine exclude guard", () => {
     );
   });
 
+  // A fixed-width lookahead read past the entry, so a workspace member on the
+  // next line marked its registry-resolved neighbour as a workspace member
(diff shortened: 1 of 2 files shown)


jan-kubica:
@coderabbitai review

PR #1663, opened 2026-08-04
fix(web): drop the row-reorder toast on table drops

jan-kubica:
## Problem

`WorkspaceTable` registered a `dropTargetForElements` on the table wrapper that accepted any `ENTITY_DRAG_TYPE` drag and unconditionally showed the `workspaces.table.reorderReadOnly` toast ("Table rows follow the current sort") on drop.

Three ways that fires when the user did not ask about row ordering:

1. **Any entity drag counts, not just table rows.** `ENTITY_DRAG_TYPE` drags start in five places (`row-cells.tsx`, sidebar `tree-view.tsx`, `kanban-card.tsx`, `calendar-entity-chip.tsx`, `overview-view.tsx`). Releasing any of them anywhere over the table produced the toast.
2. **A dropped-back row counts.** Nudging a row a few pixels and releasing, or aborting a drag over the table, hits `onDrop` with no position change.
3. **It ignored sort state**, so it claimed rows follow "the current sort" with no sort applied.

The table also has no row-reorder feature (only column reorder, via the `monitorForElements` block below it), so the message explained a capability that does not exist.

## Change

Remove the drop target. A drop on the table is now a silent no-op and the drag springs back, matching how a rejected drop behaves in kanban and calendar. Removing the last consumer of `workspaces.table.reorderReadOnly` also drops the key from `en.json`; `bun run i18n:sync` pruned it from the 13 other locales and regenerated `messages.gen.ts`.

Column reorder is untouched.

## Follow-up

Nothing now explains why dragging a row does not reorder it. There is no action to point at today, but if manual row ordering lands later, that is the point to reintroduce an explanation, gated on a sort actually being active.

## Verification

`bun run verify` passes.


jan-kubica:
@coderabbitai review

PR #1668, opened 2026-08-04
chore(release): default the maintenance-release recording attestation

jan-kubica:
## Problem

`bun run release:maintenance` aborted whenever any product-story recording was stale, which is the normal state after any UI change. Cutting a patch release therefore always required re-running with `--confirm-current-recordings-reviewed --reason "<review reason>"`, and the reason string was retyped each time.

## Change

A bare `bun run release:maintenance` now carries stale recordings forward under a standing attestation, `Patch release, UX diff negligible`, and prints which captures it covered plus the `marketing:reshoot` alternative. Passing `--confirm-current-recordings-reviewed --reason "…"` still records a specific review.

`recordingReviewReason` is now always a `string`, so the type drops `| null` and the unreachable "recordings require review" throw goes with it.

`marketing:stale --strict` in `tag-on-version-bump.yml` is unchanged, and `marketing:verify-current` still binds each attestation to a hash of the watched source tree, MP4, and poster, so a later code or media change re-invalidates it.

## Verification

`bun run verify` passes, and a bare `bun run release:maintenance` on this branch prepared 0.6.23 without prompting (reverted, not committed here).

### DIFF
--- a/apps/landing/public/media/products/README.md
+++ b/apps/landing/public/media/products/README.md
@@ -169,6 +169,14 @@ pushing a release tag, so a stale recording fails the tag push instead of
 shipping a landing page that has quietly drifted from the product. Clear it
 with `bun run marketing:reshoot` before bumping `VERSION`.
 
+`bun run release:maintenance` does not stop on stale recordings. It carries
+them forward under a standing attestation, `Patch release, UX diff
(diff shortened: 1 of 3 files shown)


jan-kubica:
patch releases have most of the time negligible ux recordings it is fine to keep as is. JK

PR #1681, opened 2026-08-04
feat(search): extract text from spreadsheets, presentations, and legacy office formats

jan-kubica:
## Problem

Text extraction covered PDF, DOCX, email, and plain-text formats. Everything else reached the search index only through the Gotenberg PDF derivative, which is a paginated rendering: `applyFitToPage` scales spreadsheet columns to fit a page, and anything past the fit never appears in the PDF at all. A spreadsheet exhibit was indexed as whatever survived that rendering.

## Change

Route spreadsheets, presentations, OpenDocument, RTF, EPUB, and legacy `.doc` through [anydoc](https://github.com/firecrawl/anydoc), which parses each format into a shared document model and serializes Markdown from it.

Formats deliberately left on their existing parsers:

- **DOCX** stays on Folio, which also recovers comments and tracked-change notes that a plain Markdown conversion drops.
- **PDF** stays on `@libpdf/core`.
- **CSV** stays on the direct-text path, where the bytes are already text.

Anything anydoc does not document (macro-enabled spreadsheets, iWork) keeps falling back to the PDF derivative.

### Extraction source selection

`pickExtractionSource` now prefers the original file whenever a parser handles its format, and falls back to the PDF derivative only for formats with no direct parser (images, HTML-to-PDF, iWork). Without this the new parsers would be unreachable for uploads, since any convertible file already has a derivative.

Two consequences worth flagging for review:

- **HTML and CSV also switch to the original.** Both are direct-text formats, so they were previously read off a LibreOffice rendering; they now go through cheerio / the text path. Better output, but a behavior change beyond office formats.
- **The derivative is still generated.** It backs the preview UI; only the extraction source changes.

### Shared MIME classification

The extraction gate and the worker each restated the same MIME-type sets, and had to agree for a format to be extractable at all: one side accepting a type the other does not handle yields a spawned subprocess that returns nothing. They now share `extractable-mime-types.ts`.

Set membership is normalized on construction. Several registered office types are spelled with capitals (`...macroEnabled.12`) while lookups are normalized to lowercase, so a literal set entry would sit there unmatchable. The disjointness test caught exactly this during development, for `.docm` and `.xlsb`.

## Safety

Parsers run in the existing sandboxed subprocess against attacker-supplied bytes, so a crash, hang, or memory blow-up costs one file's text and nothing else. A malformed-input test asserts the `null`-return contract rather than a throw.

anydoc is new (published 2026-08-04), so it is opted out of the release-age quarantine ahead of its window, listed with all seven platform subpackages per the note in `bunfig.toml` — quarantine one of them and bun skips it silently, leaving the parent with no native binding. The entry carries a `quarantine-expires` annotation and can be dropped after 2026-08-09.

## Build

The worker bundle now emits a sidecar `.node` asset beside the entry point, which `--outfile` refuses to write. The Dockerfile step switches to `--outdir`; the runtime image already copies the whole worker directory. Verified by reproducing the production bundle and running it against the fixtures.

## Tests

- Real `.xlsx` and `.pptx` fixtures through the actual worker subprocess: cell values across multiple sheets, slide titles and body text across slides.
- Malformed and truncated office files return `null` instead of throwing.
- Office documents stored as `application/octet-stream` are recovered by extension.
- Disjointness invariant over the classification sets, so dispatch order cannot silently decide which parser wins.
- `pickExtractionSource` routing, including that DOCX/XLSX/PPTX with a derivative still read the original.

`bun test src/lib/search/extract-content.test.ts` — 9 pass. `@stll/api` lint and typecheck clean.

## Not included

Widening the chat attachment allowlist (`USER_FILE_ALLOWED_MIME_TYPES`) to accept spreadsheets and decks. Same parser, but a product-facing change; happy to follow up.

## Summary by CodeRabbit

* **New Features**
  * Expanded document extraction to support additional Office, OpenDocument, EPUB, RTF, spreadsheet and presentation formats.
  * Improved recognition of files using generic MIME types or file extensions.
  * Added fallback extraction from PDF versions when the original format is unsupported.

* **Bug Fixes**
  * Improved extraction of spreadsheet cells and presentation content across files.
  * Malformed office documents now fail safely without causing extraction errors.


jan-kubica:
Took over and rebased this PR onto current `main`. Review/validation updates:

- 18 focused MIME/source-selection/Office extraction tests pass (including real XLSX/PPTX fixtures and malformed input)
- dependency audit reports no new high/critical advisories
- `bun install --frozen-lockfile` and workspace hygiene pass
- corrected the docs to call the worker process isolation, not an OS security sandbox
- removed the stray `apps/web` TypeScript 6 declaration introduced on current `main`, which was breaking workspace installs

The dependency bot’s unknown-license note is registry metadata lag: npm declares `@firecrawl/anydoc@0.1.2` as MIT and the upstream repository license API identifies the checked-in LICENSE as MIT: https://github.com/firecrawl/anydoc/blob/main/LICENSE

jan-kubica:
Took over and rebased this PR onto current `main`. Review/validation updates:

- 18 focused MIME/source-selection/Office extraction tests pass (including real XLSX/PPTX fixtures and malformed input)
- dependency audit reports no new high/critical advisories
- `bun install --frozen-lockfile`, full affected code checks, formatting, i18n, and workspace hygiene pass
- corrected the docs to call the worker process isolation, not an OS security sandbox
- removed the stray `apps/web` TypeScript 6 declaration introduced on current `main`, which was breaking workspace installs

The dependency bot’s unknown-license note is registry metadata lag: npm declares `@firecrawl/anydoc@0.1.2` as MIT and the upstream repository license API identifies the checked-in LICENSE as MIT: https://github.com/firecrawl/anydoc/blob/main/LICENSE

PR #1682, opened 2026-08-04
fix(chat): strip web-only plumbing ids so registry projections survive real payloads

jan-kubica:
## Summary

The chat registry projection's ref-field map is a hand-maintained description of each handler's payload shape. This PR adds a third lever beside ref hydration and passthrough — `stripPaths`, for fields other surfaces need but chat cannot act on — and a contract corpus that keeps the map and the handlers aligned.

- `stripPaths`: declared paths are deleted from the chat projection before hydration and the UUID backstop run. Wired into both the read and write orchestrators. Declarations added for `list_matters` (overview plumbing ids) and `list_playbooks` (tier rule/entry ids).
- The backstop's error message now names the ref projection and states the retry semantics explicitly.
- `registry-projection-contract.test.ts`: every chat-projectable read tool runs through `runRegistryReadTool` against fixtures whose every id-bearing column is UUID-formatted. Coverage is enforced by `satisfies` over the projectable-tool union, and an anti-vacuity check requires each declared outputRef path to resolve to a minted chat ref in the actual payload.

### DIFF
--- a/apps/api/src/handlers/chat/tools/registry-adapter/ref-field-map.ts
+++ b/apps/api/src/handlers/chat/tools/registry-adapter/ref-field-map.ts
@@ -101,6 +101,16 @@ export type RefMediationEntry = {
    * it, which fails closed the same as an undeclared path.
    */
   passthroughIdPaths: readonly string[];
+  /**
+   * Output paths deleted from the chat projection before hydration and the
+   * UUID backstop run. For fields other surfaces need (web-UI field/file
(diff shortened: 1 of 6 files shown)


jan-kubica:
Implemented in 8e716db: the per-call assertions now require every declared `stripPaths` entry to be absent from the projected payload (`readPathValues(...).length === 0`), so stripping is verified directly rather than inferred from the backstop's ok result.

PR #1698, opened 2026-08-05
fix: harden chat consent maps, model ingress, and lookups against silent drift

jan-kubica:
Closes a family of silent-drift gaps where a hand-maintained structure mirrored a source of truth with nothing binding them, or a lookup miss fell back silently.

## Frontend consent maps derive from the backend classification
- `@stll/api/types` now exports `BuiltInChatToolPolicyKindByName` (the per-tool policy-kind map) and `ApprovalRequiredBuiltInChatToolName` (derived type-level from the kinds whose `needsApproval` is true).
- `apps/web` derives `PublicOfficialToolName` / `ExternalInputToolName` from that map instead of hand-listing names: reclassifying a tool on the backend now breaks web typecheck instead of silently keeping stale auto-approval or consent framing.
- The `Partial<Record>` approve-once list and the bare `"spawn_subagents"` literal are replaced by one total `CHAT_TOOL_GRANT_POLICY` record over every approval-requiring built-in tool (`grantable` / `approve-once` / `never-auto`): a newly approval-gated tool cannot land without an explicit grant classification. Runtime behavior is unchanged for every current tool (pinned by tests).

## Model ingress is a typed seam covering every provider-bound surface
- The guard mints branded surfaces (`GuardedSystemPrompt`, `GuardedToolSchemas`, `GuardedModelMessages`); minting schemas are module-private, so only the guard functions can construct them. `streamChat`'s dispatch takes a single `GuardedChatSurfaces` bundle: a surface that skipped the guard, or was rebuilt after it ran, fails typecheck.
- Org MCP connectors serve tool schemas lazily; `guardMcpToolSource` wraps the source's `tools()` so every fetched batch passes the guard, composed outermost after provider projection and the third-party anonymizer.
- Runtime-middleware rewrites (compaction summaries, loop-recovery prompts) re-enter the guard before replacing config values, with the tenant id set carried inside `GuardedChatSurfaces`.
- Every model call outside `streamChat` is guarded at the `tanstack-ai-generate` choke point: `tenantWorkspaceIds` is a required parameter (callers with no tenant scope state `[]` explicitly), messages always redact, and system prompts declare `server-built` (panic on a hit) or default to redact-and-report for mixed-provenance prompts. The direct `chat()` islands (subagent runner, DOCX field generator) get the same treatment.

## Lookup misses are loud
- `getChatToolPolicy`: a policy-map miss still fails open to `internal` (org-configured MCP tools must not brick chat) but now captures deduped telemetry; tighten to fail-closed once quiet.
- Usage rate lookup: an unknown model id still bills at the fallback rate but captures deduped telemetry instead of hiding the miss.

## Tests can no longer pass vacuously
- Registry projection contract fixtures are pinned with `satisfies` against the real handler return types, so a handler growing a UUID-bearing field forces a fixture update the projection backstop can then catch. Pinning surfaced several fixture-shape mismatches (missing required fields, an invalid contact-role enum value, a `Date` where the column is a date string); all were test-only, verified against the composition code.
- The third-party boundary asserts a UUID-shaped anonymization scope id is forwarded verbatim into the anonymize call.
- The anonymize default label set is pinned against a hand-written literal oracle, so an upstream default change is a reviewed decision rather than a propagated re-export.
- The chat orientation eval now fails a task on any redaction marker or unresolved-ref sentinel in persisted output (previously notes-only).
- Every closed ingress gap carries an anti-vacuity pair with UUID-shaped fixtures: one case proving the guard fires on that surface, one proving clean input passes untouched.

## MCP egress: raw ids verified as intentional, and pinned
MCP clients are authenticated, tenant-scoped programmatic consumers whose API surface legitimately includes ids for follow-up calls; anonymization there covers declared text fields only, and the chat registry projection owns the UUID backstop on the one path that crosses into model context. This is now documented at `finalizeMcpEgress`, and the egress canary pins the positive side: ids survive anonymization byte-for-byte while declared text fields are replaced.

## Remaining limits (stated, not silent)
- Root- and org-scoped generation handlers (template fill, prompt improvement, public-corpus jobs, and similar) pass an explicit empty tenant set: the guard mechanism runs, but giving those paths a real id set would add an accessible-workspaces lookup on interactive paths, deferred deliberately.
- The MCP source's optional `readResource` is not implemented by our tool source, so nothing is served on that channel today; if it ever is, it must mint through the guard like `tools()`.

No baselines were reseeded; typecheck-instantiation and ratchet checks pass within existing headroom.

## Summary by CodeRabbit

* **Security & Privacy**
  * Improved protection against sensitive workspace information appearing in AI prompts, messages, tools and generated content.
  * Strengthened safeguards for data spanning multiple workspaces and public or organisation-scoped workflows.
  * Enhanced anonymisation checks while preserving identifiers and non-text data where appropriate.

* **Reliability**
  * Improved chat recovery, compaction and AI-generated document workflows through additional validation.
  * Added monitoring for unsupported AI models and missing tool policies.
  * Expanded benchmark validation for unresolved references and redaction markers.

* **Bug Fixes**
  * Fixed approval handling so one-time and non-persistent tool permissions are applied consistently.


jan-kubica:
@coderabbitai review

jan-kubica:
@coderabbitai review

PR #1711, opened 2026-08-05
feat(inspector): dock DOCX find in the pane

jan-kubica:
## Problem

Cmd+F over a DOCX open in the inspector opens Folio's find dialog outside the pane, on top of whatever the main page is showing.

That placement is deliberate given the dialog's design: `FindReplaceDialog` is a fixed viewport overlay positioned from `--folio-find-replace-top/left/right` and packed to the inline-end edge. `apps/web/src/routes/_protected.tsx` sets the right offset to the inspector width plus rail so the dialog clears the docked pane. In full view that keeps the dialog off the document; in sidepeek it parks it over unrelated content.

## Change

The inspector claims the shortcut and renders its own bar instead:

- `useDocxFind` listens for Cmd/Ctrl+F on `document` in the capture phase and stops propagation, so Folio's bubble-phase listener never opens the overlay. Only the bare shortcut is claimed; `classifyDocxFindKeydown` leaves every other combination alone, since suppression costs a `stopPropagation`.
- `DocxFindBar` docks at the top of the pane: query input, match counter, prev/next, close. Same shape as the find bar in `external-reference-panel.tsx`, and it reuses `SearchMatchControls`.
- Search reuses the path already behind the PDF peek viewer: `findDocumentSearchResult` plus `resolveFindMatchRange`, then `setPassageHighlight` and `scrollToPosition` for the active match. Capped at `MAX_SEARCH_PREVIEW_MATCHES`, with the counter showing `200+` when truncated.
- A new `surface` prop (`"inspector" | "fullView"`) selects the behavior. Full view keeps Folio's dialog, which has the viewport to sit in.

Background inspector tabs stay mounted but CSS-hidden, so the listener gates on `offsetParent` and only the visible editor claims the shortcut. Escape is only taken when the event target is inside the pane, leaving portalled dialogs alone.

The bar renders above Folio's formatting toolbar, since the toolbar is rendered inside `DocxEditor`.

## Notes

- No new translation keys: the `folio.findReplace.*` set already covers the bar's labels in every supported language.
- Enter inside the debounce window flushes the pending search so the first press lands on the first match; Shift+Enter steps backwards.
- Highlights the active match only, matching the existing search-preview behavior rather than Word-style all-match highlighting.

## Verification

`bun run verify` on this branch passes except "AI skill sync", which fails on any checkout without the `.ai/shared` submodule, as `AGENTS.md` notes. Typecheck, lint, and `apps/web/src/components/docx` tests are green; the one failing test in that directory (`docx-edit-mode.logic.test.ts`) fails identically on a clean tree without `apps/web/.env`.

Not yet exercised in a browser.

## Summary by CodeRabbit

* **New Features**
  * Added DOCX search in the inspector view, including keyboard shortcuts, match navigation, and a close button.
  * Search can now highlight matches in the document and jump between results.

* **Bug Fixes**
  * Improved keyboard handling so common shortcut combinations behave more consistently and Escape closes search only when appropriate.

* **Tests**
  * Added coverage for DOCX search shortcut behaviour and Escape handling.

### DIFF
--- a/apps/web/src/components/docx/docx-browser-editor.tsx
+++ b/apps/web/src/components/docx/docx-browser-editor.tsx
@@ -61,8 +61,10 @@ import {
 } from "@/components/docx-preview-zoom";
 import { DocxEditor } from "@/components/docx/app-docx-editor";
 import type { DocxComments } from "@/components/docx/app-docx-editor";
+import { DocxFindBar } from "@/components/docx/docx-find-bar";
 import { DocxLoadingShell } from "@/components/docx/docx-loading-shell";
 import { useDocxBlockScroll } from "@/components/docx/use-docx-block-scroll";
+import { useDocxFind } from "@/components/docx/use-docx-find";
 import { useFolioCollaborationSession } from "@/components/docx/use-folio-collaboration-session";
 import { useSyncDocxSuggestions } from "@/components/docx/use-sync-docx-suggestions";
 import {
@@ -122,6 +124,9 @@ const colorFromStableId = (value: string) => {
   return `#${color.toString(16).padStart(6, "0")}`;
 };
 
+/** The inspector docks the editor beside the page; full view owns the page. */
+export type DocxEditorSurface = "fullView" | "inspector";
+
 type DocxBrowserEditorBaseProps = {
   workspaceId: string;
   entityId: string;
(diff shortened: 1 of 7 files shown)


jan-kubica:
@coderabbitai review

PR #1712, opened 2026-08-05
feat(chat): orientation eval repeat runs and outcome validity

jan-kubica:
Four measurement improvements to `chat-orientation-eval.ts`:

- **Read floor**: tasks whose expected names appear in the prompt can be satisfied by echo; `minToolCalls` (set to 1 on find-document) requires the model to actually read before an affirmative counts.
- **Outcome discriminator**: `pass | fail | no-response | unsupported` replaces the boolean. Zero tool calls with empty final text is `no-response` (probable provider/adapter incompatibility, named in the notes) so it cannot be mistaken for a wrong answer.
- **Model pre-validation**: OpenRouter-shaped ids are checked against the public model catalog for tool support before any run; unsupported entries render as `unsupported` rows instead of producing vacuous task results.
- **Repeat runs**: `--runs N` with a per-(model, task) `passes/N` aggregate and `--pass-threshold` gating the exit code, so single-sample noise stops deciding red/green.

Typecheck (scripts project) and lint clean; usage header documents the flags, outcome glossary, and exit rule.

## Summary by CodeRabbit

* **New Features**
  * Added support for validating OpenRouter models before evaluation.
  * Evaluations can now run tasks repeatedly with configurable pass thresholds.
  * Added detection for unsupported models, missing responses, and insufficient tool calls.
  * Results now include per-task aggregates, outcome-based statuses, JSON reports, and pass-rate exit codes.

* **Documentation**
  * Expanded command-line options and usage guidance, including run counts and pass-threshold configuration.

### DIFF
--- a/apps/api/scripts/chat-orientation-eval.ts
+++ b/apps/api/scripts/chat-orientation-eval.ts
@@ -38,17 +38,39 @@
  *  - A bare id (no known `provider::` prefix) skips provider validation and
  *    resolves against the default provider chain.
  *
- * Flags:
- *   --api <url>         API base URL (default http://localhost:3001)
- *   --web-origin <url>  Origin header for better-auth (default http://localhost:3000)
- *   --email <email>     Sign-in email (default test@stella.dev)
- *   --models <a,b,c>    Comma-separated devModelId list; omit for one
- *                       pass with the org default model
- *   --json <path>       Write the full result object as JSON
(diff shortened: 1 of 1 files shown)


jan-kubica:
Header updated: the all-skipped pre-validation path (exit 1 with nothing to run) is now documented alongside the per-model rule.

PR #1727, opened 2026-08-05
fix(web): read identity from the provider in the docx editor

jan-kubica:
The docx editor is hosted by the inspector, which stays mounted across navigation, but it read its identity strictly from the `/_protected` route context. On a route tree that does not include `_protected` — the public law routes — that match no longer exists, so the read throws and the editor subtree falls into its error boundary until the user leaves the route. Reproducible in three steps: open a document in the inspector, open Case Law, open a decision. The read also sat above the collaboration feature check, so that flag could not gate it.

- The editor now reads `useMaybeAuthenticatedUser()`. `AuthenticatedUserProvider` wraps both the protected tree and the authenticated case-law workspace, so the identity survives navigation; the `maybe` variant exists for exactly this public-route case.
- `useFolioCollaborationSession` takes the identity as `{...} | null` and `canConnect` requires it, so a collaboration session cannot be opened without someone to publish as. Narrowing is on the name and colour primitives rather than the inline object, so the effect does not reconnect on every render.
- New `no-strict-route-read-in-chrome` oxlint rule: components that outlive their route (docx, inspector, ai-suggestions, sidebar, user menu) may not read route-scoped context strictly (`useRouteContext`, `useParams`, `useSearch`, `useMatch`, `useLoaderData`, `getRouteApi`) unless they opt out with `shouldThrow: false` / `strict: false`. The fixture asserts both directions. Scanning that scope found no other instances.

The inspector's own host component already read non-strict; the editor was the outlier.

### DIFF
--- a/.oxlint-plugins/__fixtures__/no-strict-route-read-in-chrome.fixture.tsx
+++ b/.oxlint-plugins/__fixtures__/no-strict-route-read-in-chrome.fixture.tsx
@@ -1,0 +1,36 @@
+// Passive regression fixture for
+// `no-strict-route-read-in-chrome/no-strict-route-read-in-chrome`.
+
+import {
+  getRouteApi,
+  useMatch,
+  useParams,
+  useRouteContext,
+} from "@tanstack/react-router";
+
+// Route-bound module binding in chrome — MUST flag.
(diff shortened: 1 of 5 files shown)


jan-kubica:
@coderabbitai review

PR #1738, opened 2026-08-06
feat(sidebar): hide the matter activity disclosure when it is empty

### LINKED ISSUE #1669: feat(sidebar): hide dropdown if there are no past actions to show
### Feature request

If there is nothing to show upon a click to a dropdown for past actions, then dont show the dropdown

Image

### Motivation

as a general ux principle, we should not advertise an action that leads to a dead end

### Relevant sources

_No response_

asadbek0512:
## Proposed change

Closes #1669.

A matter row's activity disclosure rendered an empty `SidebarMenuSub` when the matter had no entities and no threads, so the chevron opened onto nothing. Because the active matter auto-expands (`resolveAutomaticExpandedMatterId`), a fresh matter showed that dead end without a click.

- `MatterActivityList` returns `null` on a resolved-empty result instead of an empty list.
- `MatterItem` drops the chevron once that result is known to be empty.

Emptiness is read from the cache the expanded row already populates, with `enabled: false`, so the sidebar issues no extra requests and its network baseline is unchanged. An unresolved query stays "unknown" and keeps the disclosure available, so the toggle never disappears under the pointer.

`matterActivityIsKnownEmpty` lives in `app-sidebar.logic.ts` with cases in `app-sidebar.logic.test.ts` for the unresolved, partially populated, and fully empty page sets.

The row keeps its `ps-8` inset so a matter without a chevron stays aligned with its siblings.

## Checklist

- [x] **DARK MODE SUPPORT** | User can use the webapp in dark mode. Make sure your changes are visible in both light and dark mode.
- [x] **ISSUE LINKED** | Link an issue to this PR (not by mentioning it in the description, but by using the GitHub issue linking feature)

The change only removes UI, so there is no new surface to screenshot: the disclosure and its chevron are absent where an empty list used to render. No colours, tokens, or strings were touched, so dark mode and the translated labels are unaffected.

## Summary by CodeRabbit

- **Bug Fixes**
  - Matter navigation no longer shows an activity disclosure when activity is confirmed to be empty.
  - Empty activity sections are now hidden instead of displaying a blank submenu.
  - Activity controls remain available while results are still loading or incomplete.
- **Tests**
  - Added coverage for unknown, paginated and fully empty activity results.

### DIFF
--- a/apps/web/src/components/app-sidebar.tsx
+++ b/apps/web/src/components/app-sidebar.tsx
@@ -1,5 +1,11 @@
 import type * as React from "react";
-import { useMemo, useRef, useState } from "react";
+import {
+  useCallback,
+  useMemo,
+  useRef,
+  useState,
+  useSyncExternalStore,
+} from "react";
 
 import { combine } from "@atlaskit/pragmatic-drag-and-drop/combine";
 import {
@@ -7,7 +13,7 @@ import {
   dropTargetForElements,
 } from "@atlaskit/pragmatic-drag-and-drop/element/adapter";
 import { formatForDisplay, useHotkey } from "@tanstack/react-hotkeys";
-import { useInfiniteQuery } from "@tanstack/react-query";
+import { useInfiniteQuery, useQueryClient } from "@tanstack/react-query";
 import {
   getRouteApi,
   Link,
@@ -116,6 +122,7 @@ import { ENTITY_DRAG_TYPE } from "@/lib/workspaces/drag-constants";
 import { useUpdateWorkspace } from "@/lib/workspaces/mutations";
 import {
   workspaceActivityOptions,
+  workspacesKeys,
   workspacesNavigationOptions,
 } from "@/lib/workspaces/queries";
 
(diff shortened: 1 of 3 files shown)


asadbek0512:
I have read the CLA Document and I hereby sign the CLA

asadbek0512:
Fixed in ecdfa1d.

The current endpoint cannot actually produce an empty page with a cursor: `read-activity.ts` issues `nextCursor` only when `merged.length > limit`, and in that branch `pageItems` holds a full `limit` of items, so an empty page always carries `nextCursor: null`. The helper should not rest on that invariant, though, so it now proves the result is drained rather than assuming it:

- `matterActivityIsKnownEmpty` requires the last page to have no cursor before reporting empty.
- `MatterActivityList` returns `null` only when `items.length === 0 && !hasNextPage`, so the "show more" control stays reachable if that shape ever appears.

Added a case for a cursor-carrying empty page.

CC on behalf of asadbek0512

apps/web/src/components/app-sidebar.tsx
```diff
@@ -966,6 +968,21 @@ const MatterItem = ({
...
+  // Read whatever the activity query has already cached for this matter without
+  // ever issuing a request: the expanded row owns the fetch, so `enabled: false`
+  // keeps the sidebar at its current request count while still re-rendering the
+  // disclosure once that fetch resolves.
+  const { data: cachedActivity } = useInfiniteQuery({
+    ...workspaceActivityOptions({
+      activeOrganizationId,
+      key: { workspaceId: ws.id },
+    }),
+    enabled: false,
```
jan-kubica, line 987: A disabled TanStack Query observer does not refetch when this activity query is invalidated, so a cached empty result can keep the chevron hidden after the matter gains new activity

please treat invalidated cache data as unknown and only hide the disclosure when the cached query succeeded, is not invalidated, and contains a fully drained empty result.

jan-kubica (changes requested):

jan-kubica:
thank you for your contribution! one minor suggestion and good to go:)

asadbek0512:
Good catch, fixed in abc2f89.

Worth noting for anyone reaching for the obvious signal: the observer's own `isStale` cannot carry this. `queryObserver`'s `isStale(query, options)` short-circuits on `enabled !== false`, so a disabled observer reports `isStale: false` no matter how the entry is marked. The invalidation flag lives on the cache entry, so the row subscribes to the query cache and reads `isInvalidated` from there.

`matterActivityIsKnownEmpty` now takes `{ isInvalidated, pages, status }` and hides the disclosure only on a succeeded, non-invalidated, drained, item-free result; anything else is unknown and keeps the toggle. Added a case for the invalidated-empty result and for a failed query.

CC on behalf of asadbek0512

jan-kubica (approved):

PR #1747, opened 2026-08-06
fix(chat): draw mention icons from the entity kind, not the label text

### LINKED ISSUE #1720: bug: incorrect icon for Folders mention in Chat
### Reproduction

Folder icon expected, generic document fallback displayed
Image

### Expected behavior

Folder icon expected when mentioning folder

jan-kubica:
Closes #1720. Draws a chat mention's icon from the entity's kind rather than from its label text. Second of three stacked PRs; based on #1746.

## Change

`EntityChipIcon` in `streamdown-mention-link.tsx` selected a mention chip's glyph by testing the visible link label against `/^(?:folder|slozka|priecinok)\b/iu` and a matching task pattern. The label is the entity's name (the API mints mentions as `[<name>](#stella-entity-ref=<ws>:<id>)`), so the patterns only matched an entity literally named "Folder ..."; folders and tasks carry no file field either, so the mime branch did not apply and both fell through to the generic document glyph. The patterns also covered only English, Czech and Slovak.

The component already resolved the entity to read its mime type, and that endpoint returns `kind` (`apps/api/src/handlers/entities/get.ts`), so the kind is available in the query cache.

Four components carried their own kind-to-glyph dispatch, three of them branching on `"folder"` alone, so tasks, messages and links drew as documents in the mention picker, the source chips and the composer node view as well. All four now route through `EntityKindIcon`:

- `EntityKindIcon` becomes a `switch` over `EntityKind` with a `never` default, so a kind added to `ENTITY_KINDS` fails to typecheck here until it is given a glyph
- a new `EntityIcon` takes an `EntityIconSource` union (`resolved` / `pending` / `unknown`), so a caller that has not resolved the entity renders a neutral placeholder rather than a glyph it cannot justify
- `useEntityIconSource` resolves the entity once and is shared by the streamdown renderer and `entity-link`
- `ChatMentionOption.kind` is the closed `ChatMentionKind` union rather than `string`; untrusted TipTap attributes narrow through `isEntityKind` and yield `null`

`EntityKindIcon`'s task branch previously resolved `status ?? "open"` and drew that status's glyph. The entity endpoint does not return status, so callers without one now draw a generic task glyph instead.

`chat-mention-extension.test.ts` built its fixture as `kind: category`, a value no producer emits; the tighter type surfaced it, and the fixture now maps category to a realistic kind through a total `satisfies Record<...>` map.

## Behaviour notes for review

- Source chips: the placeholder carries its own muted tone, so resolved glyphs keep their existing styling.
- Mention picker: generic (non-Office) file glyphs now inherit the row's muted tone. The colourful file-type icons draw their own fills and are unaffected.
- The workspace overview's "New task" menu item passes `kind="task"` with no status, so it moves from the open-status circle to the generic task glyph.

## Verification

- `bun --filter @stll/web typecheck`, `lint`: clean
- `bun --filter @stll/web test`: 1494 + 54 pass, 0 fail
- `entity-kind-icon.test.tsx` asserts every kind draws a distinct glyph, that none equals the placeholder, and pins the task-status behaviour; verified red against the previous implementation

CC on behalf of jan-kubica

### DIFF
--- a/apps/web/src/components/chat-mention-extension.test.ts
+++ b/apps/web/src/components/chat-mention-extension.test.ts
@@ -1,7 +1,18 @@
 import { describe, expect, test } from "bun:test";
 
-import type { ChatMentionOption } from "@/components/chat-mention-extension";
+import type {
+  ChatMentionKind,
+  ChatMentionOption,
+} from "@/components/chat-mention-extension";
 import { selectChatSuggestionItems } from "@/components/chat-mention-extension";
+
+// A real producer never emits the category as the kind: an entity row
+// carries an entity kind, and the other categories carry their own marker.
+const KIND_BY_CATEGORY = {
+  entity: "document",
+  workspace: "workspace",
+  decision: "decision",
+} as const satisfies Record<ChatMentionOption["category"], ChatMentionKind>;
 
 const option = ({
   category = "entity",
@@ -15,7 +26,7 @@ const option = ({
   id,
   label,
   category,
-  kind: category,
+  kind: KIND_BY_CATEGORY[category],
   mimeType: null,
 });
 
--- a/apps/web/src/components/chat-mention-extension.ts
+++ b/apps/web/src/components/chat-mention-extension.ts
@@ -7,6 +7,8 @@ import {
 } from "@tiptap/react";
 import type { SuggestionOptions, SuggestionProps } from "@tiptap/suggestion";
 
+import type { EntityKind } from "@stll/api-contract";
+
 import { ChatMentionList } from "@/components/chat-mention-list";
 import { ChatMentionNode } from "@/components/chat-mention-node";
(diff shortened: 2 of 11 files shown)


jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

PR #1748, opened 2026-08-06
chore(lint): make hand-rolled entity glyphs unwritable

jan-kubica:
Keeps entity glyphs inside one component. Third of three stacked PRs; based on #1747.

## Change

`.oxlint-plugins/no-direct-entity-glyph.ts` follows the existing `no-direct-matter-glyph` rule, which keeps the matter glyph inside `MatterIcon`. It bans importing `Folder` / `FolderOpen` / `ListTodo` (with or without lucide's `Icon` suffix) from `lucide-react` outside `entity-kind-icon.tsx`.

Only the kind-bearing glyphs are covered. `FileIcon`, `FileTextIcon`, `MailIcon` and `LinkIcon` carry ordinary non-entity meanings across the app (empty states, upload surfaces, mailto buttons, hyperlinks), so the rule would produce noise rather than signal there. A passive fixture in `__fixtures__` keeps the rule honest: its disable directive goes unused if the rule stops firing, and `--report-unused-disable-directives-severity=error` fails CI.

Every remaining raw folder/task glyph in web source moves to `<EntityKindIcon>`: the file tree, the filesystem tree view and its breadcrumb and move-to-root affordances, the table folder cell, the matter folder picker, the file organizer dialog, the expand-all toggle, the todos empty state and the template category action. `EntityKindIcon` gains `folderState: "collapsed" | "expanded"` so callers owning an expand affordance keep the open-folder glyph. Each migrated site renders the same glyph with the same classes; the diffs are output-identical.

## Ratchet baseline

The new `entity-kind-glyph-adhoc` metric counts raw folder/task glyph identifiers in web source outside `entity-kind-icon.tsx`. It is seeded at **2**: the global search dialog's import and its map entry.

That file is the single entry in the rule's `ALLOWED_FILES`, because its total `Record<GlobalSearchResultType, IconType>` spans a union wider than `EntityKind` (matters, contacts, case law, chat threads) and needs component references rather than elements. The metric is what makes that exception non-growable: the lint rule alone would accept a new allowlist entry, the baseline will not. The baseline diff is the new metric's entry only; no existing count was reseeded, and no lint suppressions are added (`lint-suppression-directives` stays at 567).

Separately tracked: the search dialog draws tasks with `SquareCheckIcon` and documents with `FileTextIcon`, where the rest of the app uses `ListTodoIcon` and `FileIcon`. Unifying them changes what search results look like, so it is a product decision rather than a mechanical one.

## Verification

- `bun --filter @stll/web typecheck`, `lint`: clean
- `bun --filter @stll/web test`: 1494 + 54 pass, 0 fail
- `bun scripts/ratchet.ts --self-test`: PASS; the new counter has fixture coverage for string, comment, lowercase and longer-identifier false positives
- `bun scripts/ratchet.ts --check`: OK, 22 metrics at or below baseline
- `bash scripts/lint-oxlint-fixtures.sh`, `tsc --noEmit -p tsconfig.oxlint-plugins.json`: clean
- rule verified to fire against a scratch file importing `FolderIcon`

CC on behalf of jan-kubica


jan-kubica:
@coderabbitai review

Final head is flattened onto main; please review the standalone diff.

CC on behalf of jan-kubica

PR #1780, opened 2026-08-07
fix(reports): embed the built-in report DOCX so it resolves in the compiled server

jan-kubica:
## Proposed change

`builtin-templates.ts` located the shipped Due Diligence DOCX by URL:

```ts
const url = new URL("assets/dd-report.docx", import.meta.url);
return Buffer.from(await Bun.file(url).arrayBuffer());
```

`apps/api/Dockerfile` links the server into a standalone binary with `bun build --compile`. In that binary `import.meta.url` is `file:///$bunfs/root/server`, so the URL above resolves to `/$bunfs/root/assets/dd-report.docx`: a path that exists in the source tree but not in the binary. The Dockerfile materializes its other runtime assets under `/$bunfs/root/` (the runtime workers, the Cabinet Grotesk font, the YARA rules, the anonymize-wasm native glue, the quickjs wasm) and has no entry for this one, so `loadBuffer()` fails with `ENOENT: no such file or directory, open '/$bunfs/root/assets/dd-report.docx'`.

Both callers of `loadBuffer()` are affected, and both answer 500:

- `handlers/reports/clone-builtin.ts`, cloning the built-in into an organization's templates
- `handlers/reports/report-export-queue.ts`, filling a report export from the built-in layout

The picker is unaffected: `list-templates.ts` reads only `key` and `name` and never opens the file, so the built-in lists normally and fails at use.

This change imports the asset with `with { type: "file" }`. The bundler embeds it and rewrites the specifier to a path that resolves both when the source runs directly and inside the compiled binary, so the bytes travel with the code instead of depending on an image entry kept in sync by hand. `types/docx.d.ts` types the import.

### Verification

Compiled a binary around the real module (`bun build --compile` over an entry calling `loadBuffer()`) and ran it from `/` with no source tree present:

- before: `ENOENT ... open '/$bunfs/root/assets/dd-report.docx'`
- after: reads the 28,808-byte DOCX (`PK` magic)

`bun test apps/api/src/handlers/reports/clone-builtin.test.ts` passes (3 tests; it exercises the real `loadBuffer`), and the CI `typecheck-baseline --check` guard (API, web, and web e2e projects), focused test, scoped oxlint, and formatting checks are clean.

## Checklist

- [ ] DARK MODE SUPPORT | Not applicable: server-side asset resolution, no UI surface.
- [ ] SCREENSHOT INCLUDED | Not applicable: no user-visible UI change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SC4BbWPAFngZv6GH76DPVo)_

### DIFF
--- a/apps/api/src/docx.d.ts
+++ b/apps/api/src/docx.d.ts
@@ -1,0 +1,10 @@
+/**
+ * DOCX assets imported with `with { type: "file" }`. The import evaluates to a
+ * path the Bun runtime can read: the file's own location when the source runs
+ * directly, and a path inside the embedded asset directory once `bun build
+ * --compile` links it into the server binary.
+ */
+declare module "*.docx" {
+  const path: string;
+  export default path;
+}
--- a/apps/api/src/handlers/reports/builtin-templates.ts
+++ b/apps/api/src/handlers/reports/builtin-templates.ts
@@ -10,9 +10,21 @@
  * The manifest is embedded in the DOCX (customXml) by the generation script, so
  * the fill pipeline reads it from the buffer; it is also exported here for the
  * picker read surface and the generation script.
+ *
+ * The asset is pulled in through a `with { type: "file" }` import rather than
+ * resolved from `import.meta.url`. `bun build --compile` (see
+ * `apps/api/Dockerfile`) links the server into a standalone binary whose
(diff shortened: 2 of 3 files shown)


jan-kubica:
@codex review

PR #1840, opened 2026-08-09
fix: skip browser jobs on registry rate limits

jan-kubica:
## Summary

- classify the explicit Docker Hub pull-rate-limit response as an unavailable E2E stack
- skip dependent browser steps while preserving a visible workflow warning and summary
- keep every other stack or test failure blocking
- enforce the downstream step guards with a workflow invariant test

CC on behalf of jan-kubica


jan-kubica:
@codex review

CC on behalf of jan-kubica

PR #1908, opened 2026-08-11
fix: prevent stale playbook approvals

jan-kubica:
## Summary

- track Playbook editor dirty state and guard in-app, browser, and back navigation
- require saving before approval and reject stale multi-editor approval requests
- reuse shared localized leave-confirmation copy across Playbooks, Templates, and Clauses

## Validation

- API and web typechecks
- focused type-aware lint
- Playbook editor unit tests
- Playbook approval/version integration suite
- i18n consistency checks


jan-kubica:
Rebased onto main and reworked per a full review: the concurrency guard now covers the destructive update path as well as approve (shared `assertUnchangedSince` helper, also adopted by the style-set editor; machine-readable `version_conflict` code shared through the api contract), the editor recovers from a 409 by resyncing from the refreshed detail instead of looping, the create-form no longer reports a pristine draft as dirty, the navigation-blocker dialog no longer dead-ends on a failed save, the disabled-approve explanation actually renders, dirty tracking fingerprints the real save payload (drift now fails typecheck), and the capability catalog/CLI artifacts are regenerated for the new bodies.

PR #1911, opened 2026-08-11
fix(api): derive entity kind request schemas from ENTITY_KINDS

jan-kubica:
`packages/api-contract/src/entity-kinds.ts` is the documented single source for entity kinds, but three handlers hand-listed the literals in their `excludedKinds` schemas, with `maxItems: 5` as a second copy of the list length. A new entity kind would have been accepted by the column enum and produced by the web filters, then rejected by these validators.

`entities/list.ts`, `entities/read-window.ts`, and `entities/read-kanban-group.ts` now use `t.UnionEnum([...ENTITY_KINDS])` with `maxItems: ENTITY_KINDS.length`, matching `flows/schema.ts`. The generated CLI route map is byte-identical (codegen run, no diff), since members and order are unchanged.

Follow-up worth considering: `search/search.ts` expresses the same field via the shared `entityKindSchema` from `db/schema-validators.ts`. Both shapes derive from `ENTITY_KINDS`, so neither can drift, but collapsing the four call sites onto the shared schema would leave one idiom.

### DIFF
--- a/apps/api/src/handlers/entities/list.ts
+++ b/apps/api/src/handlers/entities/list.ts
@@ -1,6 +1,8 @@
 import { panic, Result } from "better-result";
 import { t } from "elysia";
 
+import { ENTITY_KINDS } from "@stll/api-contract";
+
 import { createSafeHandler } from "@/api/lib/api-handlers";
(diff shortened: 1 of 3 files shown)


jan-kubica:
@coderabbitai review

PR #1942, opened 2026-08-11
feat(case-law): adopt publisher document identities and correct page origins

jan-kubica:
Four sources stored rows keyed only on `(caseNumber, language)`, so distinct publisher documents sharing a docket collapsed into one row (measured instances: 19 DMS documents under one plenary docket; judgment and order under one CJEU number). This adopts each publisher's own document identity, with one shared mechanism and one migration story.

**Commit 1 — publisher document identities** (sk-us: DMS documentId; eu-ecj: CELEX+language; cz-ns: Domino unid; cz-nss: portal documentId). Each id flows through the source's single exported identity rule, used by the crawl and the reconciliation listing alike, and is bounded by `isPersistableSourceDocumentId` at the publisher boundary. Existing rows migrate in place: every build path passes the row's deterministic legacy URL via `legacySourceUrls`, so a re-observed docket-keyed row gains the id (same row, no duplicate) while a second document under the same docket becomes a second row. A table-driven db test proves the invariant per source through the real `processDecision` — including the ECLI-less case, which is what shows the URL hint alone carries the migration — and was verified non-vacuous (removing one source's hint fails its cases with a duplicate insert).

No bulk backfill: post-deploy, the reconciliation loop reads not-yet-re-keyed rows as missing and re-fetches each document once, which re-keys it; the settled-slice recheck walks all history on its standing cadence. This is a bounded, rate-limited, one-time re-observation per source, stated here deliberately as the expected traffic profile.

The eu-ecj `shared_identity` event is removed with its tests: it existed to measure this collapse, and under the new key it could only report zero.

**Commit 2 — page-origin correctness.** The sk-courts endpoint is 1-indexed and clamps below (verified live), while the crawl's paginated fetch was configured zero-indexed: every traversal re-read its first page and the live window covered 4,900 of its intended 5,000 items. Fixed at the class level: `zeroIndexed?: boolean` is replaced by a required `firstPage: 0 | 1` each adapter states exactly once, shared by its crawl and its listing walk — the defect was precisely that the two stated different origins. Persisted cursors keep their meaning (item offsets); the fix changes only the page computed from them, the crawl's position advances by exactly one page at cutover, and nothing re-walks. The single 100-item page a mid-run backfill walk would skip at the boundary is refilled by this source's per-date reconciliation ledger, which enumerates decision dates independently of the cursor. Tests drive the real adapter against a clamping stub (verified non-vacuous: under the old origin they fail, one exactly at 4,900 vs 5,000) and the conformance stub was corrected — it had modelled the bug.

Suites: adapters 214 green, postgres-gated ingestion 78 green, whole case-law handler tree 963 green. Ratchet and typecheck baselines untouched.

### DIFF
--- a/apps/api/src/handlers/case-law/ingestion/adapters/at-courts.ts
+++ b/apps/api/src/handlers/case-law/ingestion/adapters/at-courts.ts
@@ -44,6 +44,13 @@ const RIS_DOCUMENT_HOST_POLICY = {
 /** RIS API enforces a maximum page size of 20. */
 const PAGE_SIZE = 20;
 
+/**
+ * RIS numbers `Seitennummer` from one, which is also what the count probe
+ * below asks for. Stated once, so the crawl's page arithmetic and every direct
+ * request agree about where the collection starts.
+ */
+const FIRST_PAGE = 1;
+
 type RisContentUrl = {
   DataType?: string | null;
   Url?: string | null;
@@ -444,7 +451,7 @@ export const atCourtsAdapter = defineSourceAdapter({
     try {
       const response = await fetchWithTimeout(
         `${API_URL}?${new URLSearchParams({
-          Seitennummer: "1",
+          Seitennummer: String(FIRST_PAGE),
           Seitengroesse: String(PAGE_SIZE),
           Sortierung: "Aenderungsdatum",
           Aufsteigend: "false",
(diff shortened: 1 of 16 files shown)


jan-kubica:
@coderabbitai review

jan-kubica:
@coderabbitai review

PR #1964, opened 2026-08-12
fix(web): model provider validation states explicitly

jan-kubica:
## Summary

- model onboarding provider validation as a discriminated union
- require validation fingerprints for active result states and forbid them for idle rows
- make provider preview projection exhaustive

CC on behalf of artichoke


jan-kubica:
@coderabbitai review

CC on behalf of artichoke

PR #1967, opened 2026-08-12
fix(api): preserve document review topic order

jan-kubica:
## Summary

- normalize findings in confirmed-topic order instead of provider response order
- retain the first valid finding per topic while ignoring duplicates and unknown topics
- keep missing topics represented as not comparable

CC on behalf of artichoke


jan-kubica:
@coderabbitai review

CC on behalf of artichoke

PR #1974, opened 2026-08-12
fix(api): durable search projection for duplicated and copied entities

jan-kubica:
Matter duplication and entity copy flows indexed their copies only through a post-commit `processExtraction(...).catch(captureError)`. For documents that is repaired eventually (the extraction recovery scan re-creates lost runs), but everything routed through the helper's other branch — tasks, events, notes, folders, encrypted and non-extractable files, i.e. most of a duplicated matter — had no backstop: a crash between commit and the call left those copies permanently unindexed.

**Ownership discriminator** (`SEARCH_INDEX_OWNER` + `searchIndexOwnerForFields` in `process-extraction.ts`): each copy is owned either by durable extraction (the worker indexes after extracting) or by a search mark. `processExtraction` reads the same predicate for its own branch, so the handler-side split cannot drift from what extraction actually does. Mark-owned copies get `enqueueEntitySearchRepairs` INSIDE the copying transaction (rolls back with the copies; drained by the standing scheduler task), with the post-commit flush and `processExtraction` as acceleration only, and no copy is indexed twice. The split lives once in `copyEntities`, so `entities/duplicate`, `entities/copy-to-workspace`, and any future caller inherit it; `CopyEntitiesSuccessResult` now returns ids grouped by owner (the flat list was a duplicate projection; response shapes unchanged).

Also fixed in passing, same files: source field reads now carry the `orderBy id asc` + properties bound that `findExtractionFileField` documents as mandatory (an unordered read could pick a different extraction source for the copy than the source entity used), and the matter-duplicate entity-cap check moved before the write transaction — it previously returned a failure from inside the callback, committing the partial matter shell while answering 400 and deleting the copied objects the rows pointed at.

PGlite tests pin the ownership partition in both directions, in-tx rollback of marks, drain-once semantics, and per-handler mark/extraction routing. Two partial `process-extraction` module mocks now spread the real module.

Follow-ups noted, not taken here: `copyEntities` has three mid-loop `{ ok: false }` returns with the same commit-on-failure shape (separate focused PR), and writing the extraction run row inside the caller's transaction would require relaxing the deliberate insert-block policy on `document_processing_runs`, which deserves its own review.

### DIFF
--- a/apps/api/src/handlers/entities/copy-to-workspace.test.ts
+++ b/apps/api/src/handlers/entities/copy-to-workspace.test.ts
@@ -41,9 +41,36 @@ void mock.module("@/api/lib/s3", () => ({
   },
 }));
 
-const processExtractionMock = mock(async () => {});
+// Spread the real module: the copy helper reads its search-index ownership
+// split from here. Only the durable extraction request itself is replaced.
+const realProcessExtraction =
+  await import("@/api/lib/search/process-extraction");
+const processExtractionMock = mock(async (_entityId: string) => undefined);
 void mock.module("@/api/lib/search/process-extraction", () => ({
+  ...realProcessExtraction,
   processExtraction: processExtractionMock,
+}));
+
+const enqueueEntitySearchRepairsMock = mock(
+  async (_tx: unknown, _entityIds: readonly string[]) => undefined,
+);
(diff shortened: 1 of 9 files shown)


jan-kubica:
@coderabbitai review

PR #1995, opened 2026-08-12
chore(test): spread real modules in partial capture mocks

jan-kubica:
The export-dropping mock baseline (decrease-only, seeded at 86 by the module-mock factory guard) had its largest cluster on `@/api/lib/analytics/capture`: 24 files replacing the module with only the exports they override, so any co-batched consumer of a dropped export breaks by bin-packing rather than by cause.

28 factories across 28 files now spread the real module (`const real = await import(...)` + `...real`, the documented pattern): the full capture cluster, one `analytics/client` mock, and three `flow-run-events` mocks that were hand-mirroring a constant. Selection criteria for what was safe to make real: the module's top level declares only constants and functions with clients constructed lazily; no transitive import of `@/api/db/root` or a queue/network client the suite deliberately mocks away; and the module is already imported for real elsewhere in the test tree. Everything failing those criteria stays counted, by named reason (live-pool transitives, auth/crypto/network surfaces, AI provider registries).

`EXPORT_DROPPING_MOCK_BASELINE` drops 86 → 55, exact with no slack: the guard now fails immediately on any new partial mock, naming offenders (probed at 54, fails listing all 55; restored, green). All 28 touched suites plus the three mock-guard suites pass individually under isolation with zero behavioral fixes needed — every suite already overrode the export it controls, so the real siblings are inert. The batching interaction was checked: the added import cannot narrow batches because the batcher already excludes co-residents of a mocked module.

Remaining largest cluster (`process-extraction`, 7 files) needs a db-handle injection refactor rather than a mock-shape change; recorded as its own follow-up.

### DIFF
--- a/apps/api/src/handlers/chat/stream-chat-model-ingress.test.ts
+++ b/apps/api/src/handlers/chat/stream-chat-model-ingress.test.ts
@@ -4,7 +4,9 @@ import { describe, expect, mock, test } from "bun:test";
 import { toSafeId } from "@/api/lib/branded-types";
 
 const captureErrorMock = mock();
+const realCapture = await import("@/api/lib/analytics/capture");
 void mock.module("@/api/lib/analytics/capture", () => ({
+  ...realCapture,
   captureError: captureErrorMock,
   captureRequestError: captureErrorMock,
 }));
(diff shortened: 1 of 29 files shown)


jan-kubica:
@coderabbitai review

PR #2077, opened 2026-08-14
fix: wait for prompts in case-law capture

jan-kubica:
The nightly case-law capture could record the chat logo before saved-skill suggestions were ready. Fresh marketing sessions now seed the default skills, and the capture waits for the suggestion component inside the one visible inspector panel.

The explicit snapshot-update command now uses `--update-snapshots=all`; the ordinary comparison keeps its 3% cross-platform rendering tolerance, but an intentional refresh can no longer pass while preserving a stale golden. Light and dark `story-public-data-3` assets now show all four prompt cards.

Validation:
- targeted marketing screenshot update
- targeted strict marketing screenshot comparison
- affected landing/web lint and typecheck
- formatting and i18n checks

CC on behalf of jan-kubica


jan-kubica:
Addressed on `fb84612c2e`: fresh capture sessions seed the default skills, then wait for the prompt-suggestions component inside the single visible inspector panel. The explicit update command now rewrites all snapshots, so the 3% comparison tolerance cannot preserve a stale blank golden. Both regenerated assets show the four prompt cards, and the strict non-update marketing screenshot test passes.

CC on behalf of jan-kubica

PR #2099, opened 2026-08-14
fix(db): accept the rewritten time-entry-timer migration history

jan-kubica:
The 20260809120002 migration was amended during review after development databases had applied an earlier revision, so those databases fail the schema-drift check on a hash that no longer exists in code. Registers the migration in rewrittenMigrationHistories with its current hash and the ten prior revisions, following the existing entry's pattern. requiredIndexes stays empty deliberately: the unique index this migration builds is already registered in ONLINE_MIGRATION_INDEXES with createSql, and restating it would register the same name twice and confuse the name-keyed test harness. The property test re-hashes migration.sql against currentHash, so the entry is self-verifying.

## Summary by CodeRabbit

* **Chores**
  * Updated migration history metadata to support revised migration records.
  * No changes to end-user functionality or required database indexes.

### DIFF
--- a/apps/api/src/lib/db/migration-history.ts
+++ b/apps/api/src/lib/db/migration-history.ts
@@ -283,6 +283,29 @@ const rewrittenMigrationHistories = {
     ],
     requiredIndexes: [],
   },
+  // The file was amended after databases had already applied it: review
+  // tightened the legacy-timer repair to skip owners without current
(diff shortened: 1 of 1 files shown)


jan-kubica:
Superseded: not needed beyond a one-time local environment repair.

PR #2112, opened 2026-08-15
feat(api): fold diacritics in the corpus index doc mapping

jan-kubica:
## What

Two changes to the corpus index, ahead of the next index generation.

**1. Fold diacritics in the doc mapping.** The corpus is Czech and Slovak
legal text, written with diacritics. The built-in `default` tokenizer only
lower-cases, so `skody` did not reach `škody` and `Nejvyssi` did not reach
`Nejvyšší`: readers who type without diacritics (a very common input mode)
silently got fewer results, with no signal that anything was missed.

The doc mapping now declares a custom tokenizer

```json
{ "name": "folded", "type": "simple",
  "filters": ["lower_caser", "ascii_folding", "remove_long"] }
```

and the three full-text fields (`title`, `text`, `heading_path`) use it
instead of `default`. `ascii_folding` runs over both indexed terms and query
terms, so the match is symmetric: neither the corpus nor the query has to be
normalized by hand, and a query *with* diacritics keeps working unchanged.
`remove_long` drops tokens past the default byte limit, keeping OCR runs out
of the term dictionary; `lower_caser` preserves what `default` already did.

Folding is lossy by design (it cannot tell `s` from `š`), which is the right
trade for recall-first retrieval where the API reranks afterwards. Raw-tokenized
identifier fields (`document_id`, `ecli`, `court`, …) are untouched, so exact
lookups keep their exact semantics.

**2. Align the dev engine with the deployed one.** The compose service ran
`0.9.0` while deployments run `0.8.2`, so a doc mapping validated locally was
validated against a different engine than the one that has to accept it. Now
pinned to `0.8.2` by multi-arch manifest list digest, matching how every other
compose service is pinned; `linux/amd64` and `linux/arm64` both resolve from it.

## Safe to land ahead of the rebuild

`ensureIndex` (`apps/api/src/lib/corpus-index/core.ts`) probes `indexExists`
and calls `createIndex` **only when the index is absent**. It never diffs,
patches, or re-applies the config of an index that already exists, and there
is no other caller that updates a doc mapping. So this change is inert for
every currently deployed index: they keep the mapping they were built with.
It takes effect when the next generation is created and backfilled, which is
the only point at which a doc-mapping change can take effect at all (the
engine cannot retokenize existing splits).

## Validation

The tokenizer JSON was checked against a real engine at the deployed version
(0.8.2), not just typechecked: a throwaway container was given the exact
config the code emits, then three Czech/Slovak documents were ingested.

```
== search: skody (no diacritics) should match 'škody' ==
num_hits: 2
 hit: doc-3 | Súd priznal náhradu škody, ktorá vznikla porušením povinnosti.
 hit: doc-1 | Žalobce se domáhal náhrady škody způsobené nesprávným úředním postupem.

== search: škody (with diacritics) ==
num_hits: 2
 hit: doc-3 | Súd priznal náhradu škody, ktorá vznikla porušením povinnosti.
 hit: doc-1 | Žalobce se domáhal náhrady škody způsobené nesprávným úředním postupem.

== search: NEJVYSSI (uppercase, folded) should match 'Nejvyšší' ==
num_hits: 1
 hit: doc-1 | Nejvyšší soud – náhrada škody
```

The engine also echoed the `tokenizers` block back verbatim in the stored
index config, so the shape is accepted as written rather than silently
defaulted.

## Guards

`CorpusIndexFieldMapping["tokenizer"]` is typed from the declared tokenizer
list, so a field cannot name a tokenizer the doc mapping does not ship (the
engine rejects that at index creation; now it fails at typecheck instead).
A test asserts declared and used tokenizer sets are equal **in both
directions**, and that every position-recorded field folds, so a new
full-text field added with `default` trips the suite rather than shipping a
half-folded mapping.

## Summary by CodeRabbit

* **New Features**
  * Improved legal search matching across titles, text and heading paths by normalising capitalisation and accented characters.
  * Improved handling of unusually long search terms for more reliable results.

* **Bug Fixes**
  * Standardised text analysis across searchable fields, helping produce more consistent search results.

* **Compatibility**
  * Updated the search service configuration to maintain reliable operation across supported deployment environments.

### DIFF
--- a/apps/api/src/lib/legal-search/corpus-index-config.test.ts
+++ b/apps/api/src/lib/legal-search/corpus-index-config.test.ts
@@ -51,7 +51,7 @@ test("no field repeated across a document's passages is free-text searchable", (
   expect(
     config.doc_mapping.field_mappings.find((f) => f.name === "heading_path")
       ?.tokenizer,
-  ).toBe("default");
+  ).toBe("folded");
 
   // Anything else a default search field reaches must be per-passage content
   // (`text`) or written once per document (`title`).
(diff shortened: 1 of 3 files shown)


jan-kubica:
@codex review

PR #2113, opened 2026-08-15
feat(api): read quoted spans as corpus-index phrase queries

jan-kubica:
## Phrase search

Free text reached the corpus index as individually quoted words ANDed together
(`("náhrada" AND "škody")`), so a reader asking for a phrase got every decision
mentioning both words anywhere in the text. `title` and `text` are indexed with
`record: position`, so the engine can answer the adjacency question directly.

A balanced quoted span now becomes a single phrase clause; the remaining words
keep the term treatment they had. Measured against the Czech case-law index:

| query | hits |
| --- | --- |
| `("náhrada" AND "škody")` | 21,087 |
| `("náhrada škody")` | 10,696 |

### Quote conventions

The corpus is written in several typographic conventions, so the parser accepts
the pairs its own documents use rather than only the English one:

| pair | convention |
| --- | --- |
| `"…"` | straight |
| `„…“` | Czech, German |
| `„…”` | Polish |
| `“…”` | English |
| `«…»` | French, Russian |
| `»…«` | German |

Single quotes are deliberately excluded: an apostrophe inside a word (`l'État`,
`d'une`) would open a span that never closes and swallow the rest of the query.
`“` is both an English opener and a Czech closer, which works because a
character is read as an opener only where it starts a span.

### Safety

A phrase is no more expressive than a term was. The span's content goes through
the same reduction a loose term does — tokenized to unicode word characters,
then quoted with `quoteCorpusValue` — so field clauses, booleans, wildcards,
parentheses and escapes cannot survive into the clause. Degradation is
deliberate rather than incidental:

- an unbalanced quote carries no span, and its text stays the terms it would
  have produced without the quote; never an engine parse error
- an empty or whitespace-only span contributes no clause at all
- quote-free input yields byte-identical clauses to before, pinned by test

Tests cover quoted, unquoted, mixed, multiple phrases, unbalanced and empty
spans, each quote convention, Czech diacritics, Arabic script, and injection
attempts carrying quotes, backslashes, field clauses and boolean operators.

## One query assembler

The two corpus-index read paths assembled their queries separately and had
already drifted apart: the public search handler used the shared clause builder,
while `corpus-index-provider.ts` kept a private `buildQuery`/`quote` pair that
interpolated its free text straight into the DSL and escaped quotes without
first escaping backslashes (so a trailing backslash in a filter value could
swallow its own closing quote).

Both paths now call one `caseLawCorpusQuery(text, filters)`. That leaves a
single escaping rule to audit, and it is why phrase support reaches both
without a second implementation. A guard test in
`corpus-index-query-guard.test.ts` fails the build if either path grows a local
quoting helper, a field clause of its own, or a template literal around user
text again.

## Tokenizer shape

The quoted-span split is exported as one `tokenizeCorpusFreeText` returning a
`{ type: "phrase" | "term"; value }` union, consumed by `corpusFreeTextClause`
through a switch with a `never` exhaustiveness check. A term may be rewritten
where a phrase may not, so a future rewriting pass reads the same token stream
rather than re-deriving the phrase boundary from the raw string.

## Not included: chronological browse from the index

The second half of the intended scope — serving the no-query decisions browse
from the corpus index, sorted by `decision_date` — is not here. It needs a
stable secondary sort key the engine does not have:

```
POST /api/v1/case_law_v2_cze/search  {"query":"seq:0","sort_by":"decision_date,document_id"}
-> Invalid argument: sort by field on type text is currently not supported `document_id`
```

`document_id` is a fast field but a text one, and the remaining numeric fast
fields (`year`, `seq`, `citation_count`, `citation_authority`) are not unique,
so a two-key sort cannot produce a stable tiebreak. Without one, a keyset
cursor over an offset-paged engine cannot guarantee that a page boundary lands
in the same place on the next request. Details and the measurements behind that
conclusion are in the PR discussion; the browse keeps its current SQL path.

## Summary by CodeRabbit

* **New Features**
  * Improved case-law search handling for quoted phrases, multilingual text, punctuation, whitespace and Unicode characters.
  * Added support for combining free-text searches with court, date, document type, language and source filters.
  * Malformed or empty quoted phrases are handled safely without disrupting search.

* **Bug Fixes**
  * Strengthened query handling to prevent unsafe input from affecting search behaviour.
  * Standardised query construction across public case-law search experiences.

### DIFF
--- a/apps/api/src/handlers/case-law/corpus-index-query-guard.test.ts
+++ b/apps/api/src/handlers/case-law/corpus-index-query-guard.test.ts
@@ -447,6 +447,32 @@ test("the ingestion role can take every lock the corpus checkpoint holds", async
   }
 });
 
+// One assembler, or the two boundaries drift into two answers about what the
+// engine sees. The public search path and the shared provider must both reach
+// the engine through corpus-query.ts, and neither may re-derive quoting or
+// interpolate user text into the DSL itself.
+test("every case-law corpus query is assembled by the shared builder", async () => {
+  const consumers = await Promise.all(
+    [publicSearchSource, sharedSearchProviderSource].map(
+      async (source) => await Bun.file(source).text(),
+    ),
+  );
+  for (const source of consumers) {
+    expect(source).toContain(
+      'import { caseLawCorpusQuery } from "@/api/lib/legal-search/corpus-query"',
+    );
+    expect(source).toContain("caseLawCorpusQuery(");
+    // A local escaping helper is how the two paths diverged before.
+    expect(source).not.toMatch(/const quote = /u);
+    expect(source).not.toMatch(/replaceAll\('"'/u);
+    // No field clause is assembled outside the shared builder, and free text
+    // never reaches a template literal.
+    expect(source).not.toMatch(/`(?:document_type|source|language|court):/u);
+    expect(source).not.toMatch(/`decision_date:\[/u);
+    expect(source).not.toMatch(/\$\{\w+\.query\}/u);
+  }
+});
+
 test("every case-law corpus search boundary uses generation projection state", async () => {
   const consumers = await Promise.all(
     [publicSearchSource, sharedSearchProviderSource].map(
--- a/apps/api/src/handlers/case-law/decisions/search.ts
+++ b/apps/api/src/handlers/case-law/decisions/search.ts
@@ -32,10 +32,7 @@ import {
 } from "@/api/lib/legal-search/case-law-corpus-projection";
 import { corpusGeneration } from "@/api/lib/legal-search/corpus-family";
 import { readCorpusIndexSearchPage } from "@/api/lib/legal-search/corpus-index-pagination";
-import {
-  corpusFreeTextClause,
-  quoteCorpusValue,
-} from "@/api/lib/legal-search/corpus-query";
+import { caseLawCorpusQuery } from "@/api/lib/legal-search/corpus-query";
 import { loadFtsSearchConfigs } from "@/api/lib/legal-search/fts-config";
 import {
   corpusIndexId,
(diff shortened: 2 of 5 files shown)


jan-kubica:
Measurements behind the scope note at the end of the description, all against
the Czech case-law index (1,273,697 documents at `seq:0`).

**No stable secondary sort key.** A `sort_by` naming `document_id` in either
position is rejected:

```
{"query":"seq:0","sort_by":"decision_date,document_id"}
{"query":"seq:0","sort_by":"-decision_date,-document_id"}
-> Invalid argument: sort by field on type text is currently not supported `document_id`
```

`document_id` is `fast: true` but text-typed. The other fast fields on the
case-law mapping are `year`, `seq`, `citation_count` and `citation_authority`,
none unique per document, so no two-key sort yields a total order. A keyset
cursor over an offset-paged engine needs one: without it, two documents sharing
a `decision_date` have no defined order, and a page boundary that falls between
them is not reproducible on the follow-up request.

**Sort direction is worth pinning before anyone relies on it.** On this engine
version an unprefixed field and a `+`-prefixed field both sort descending, and
`-` sorts ascending:

| `sort_by` | first three `decision_date` values |
| --- | --- |
| `decision_date` | 2244-06-13, 2235-06-14, 2216-06-13 |
| `+decision_date` | 2244-06-13, 2235-06-14, 2216-06-13 |
| `-decision_date` | 1682-04-30, 1682-04-30, 1682-04-30 |

**Tie width, for whoever picks this up.** Grouping `seq:0` by day over
2010-2026 gives 4,680 non-empty days with a maximum of 2,078 documents on one
date, so an API-side tiebreak would have to re-read up to ~2.1k engine hits per
page to resume deterministically. That is inside the existing scan cap but well
above what the current SQL browse costs per page, so it wants a deliberate
decision rather than a drive-by change.

**Missing dates.** 1,273,455 of 1,273,697 documents match
`decision_date:[* TO *]`; the 242 without a date match no range clause at all
and would need an explicit placement rather than falling out of the listing.

jan-kubica:
@codex review

PR #2124, opened 2026-08-16
fix(search): key morphological expansion with the shared ASCII fold

jan-kubica:
## Problem

`foldExpansionKey` is the one definition the offline dictionary builder and the
query-time loader share: it turns a surface form into the key both sides look
up. It folded with `stripDiacritics`, which is NFD plus combining-mark removal,
and that can only fold a letter with a canonical decomposition.

Polish `ł` has none. So the key kept it (`wyłączenie` -> `wyłaczenie`) while the
corpus index's `ascii_folding` and PostgreSQL `unaccent()` both store `l`
(`wylaczenie`). Every Polish key carrying `ł` therefore addressed a spelling no
indexed lexeme has, and was unreachable from a reader typing ASCII.

`foldToAscii` (`packages/text-normalize/src/ascii-fold.ts`) is the fold already
used by the query and highlight paths, generated from the extension's own rules
and pinned against a live PostgreSQL by
`apps/api/src/lib/search/ascii-fold-unaccent-parity.db.test.ts`.

## Change

- `foldExpansionKey` now returns `foldToAscii(term.toLowerCase())`. The fold
  decomposes internally, so the separate NFC pass is gone; the existing
  NFD/NFC agreement test pins that the key stays form-independent.
- The case-then-fold order mirrors the tokenizer, which lists `lower_caser`
  before `ascii_folding`, and the builder's vocabulary arrives already
  lowercased from `to_tsvector('simple', ...)`. The orders are not
  interchangeable for a character whose lowercase form the fold table does not
  reach: `Ɩ` folds to `I` but lowercases to `ɩ`, so folding first would key an
  uppercase query on `iota` against a corpus form keyed `ɩota`. The agreement
  is now bound rather than documented: the test reads `FOLDED_TOKENIZER`'s own
  filter list, so reordering the filters fails there.
- Comments updated where they described the old mark-strip behaviour.

Call sites checked: the builder (`apps/api/src/scripts/build-expansion-dictionary.ts`)
imports `foldExpansionKey` rather than carrying its own copy, and the loader and
`expandTermWith` do the same, so builder and loader still fold identically by
construction. The stem path is unchanged: `morphology/stem.ts` performs no fold
at all (it only normalises to NFC and lowercases, because the suffix tables are
diacritic-bearing), and the fold-after-stem ordering is untouched.

## Ambiguity

Folding to ASCII widens the ambiguous-key set rather than narrowing it: `łaska`
and `laska` are distinct Polish words the index stores as one lexeme, so they
now claim one key. That is the same ambiguity as Czech `rada`/`řada`, and takes
the same established answer: the key expands to nothing, because an ambiguous
lookup is resolved by declining to answer, never by serialization order. Under
the mark strip the pair kept separate keys and each expanded to the other
family's inflections, which the index could not have matched anyway.

## Dictionaries

No dictionary has been published yet, so there is no re-key migration and no
stale payload to invalidate. Dictionaries must be built after this lands; a
payload built with the old fold would key Polish buckets in a spelling the
loader no longer produces.

## Tests

- `foldExpansionKey` fold test extended with the `ł`, `zażółć`, and `Straße`
  cases, plus a separate test binding the fold order to the tokenizer's
  filter list.
- New parser test pinning that the `łaska`/`laska` pair is detected as
  ambiguous and drops both keys.
- New `expandTermWith` test pinning that a Polish term typed without `ł`
  reaches its bucket. Both new tests fail against the previous fold.
- Round-trip and clause-safety properties are unchanged and still pass.

`apps/api` legal-search suites plus the module-mock guards: 248 tests, 0 fail.
`@stll/text-normalize`: 80 tests, 0 fail. tsc, type-aware oxlint, ratchet, and
oxfmt clean.

### DIFF
--- a/apps/api/src/lib/legal-search/expansion.test.ts
+++ b/apps/api/src/lib/legal-search/expansion.test.ts
@@ -55,6 +55,22 @@ test("an unaccented spelling reaches the accented bucket", () => {
   ]);
 });
 
+// `ł` has no canonical decomposition, so a mark-stripping fold left the key
+// spelled `wyłaczenie`: unreachable from an ASCII keyboard, and spelled unlike
+// any lexeme the index stored. The key has to fold the way the index folded.
+test("a Polish term typed without ł reaches its bucket", () => {
+  const entries = dictionaryOf([
+    {
+      documentFrequency: 600,
+      forms: ["wyłączenie", "wyłączenia"],
+      stem: "wyłącz",
+    },
+  ]);
+  expect(entries.get("wylaczenie")).toBe("wyłączenie,wyłączenia");
+  expect(expandTermWith(entries, "wylaczenie")).toEqual(["wyłączenia"]);
+  expect(expandTermWith(entries, "wyłączenie")).toEqual(["wyłączenia"]);
+});
+
 // Pinned from corpus data: `veci` and `vek`/`veku` share a stem but not a
 // word. Three shared leading characters is what separates them.
 test("a term never expands to a bucket member it shares under three characters with", () => {
--- a/apps/api/src/lib/legal-search/expansion.ts
+++ b/apps/api/src/lib/legal-search/expansion.ts
@@ -439,10 +439,10 @@ export const expandTermWith = (
   entries: ReadonlyMap<string, string>,
   term: string,
 ): readonly string[] => {
-  // Fold first, then test the shape. The fold canonicalises and strips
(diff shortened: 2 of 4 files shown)


jan-kubica:
@codex review

jan-kubica:
@codex review

jan-kubica:
@codex review

jan-kubica:
@codex review

PR #2131, opened 2026-08-16
fix(e2e): make network-baseline waterfall depth load-monotone

jan-kubica:
A `TabsList` wider than its container clipped the overflowing tabs with no way to
reach them: the list had no overflow handling at all. Fixing it in the primitive
means every consumer inherits the affordance instead of diverging.

## The list is the scroll container

`overflow-auto` goes on the list element itself, not a wrapper, for two reasons
that only hold there:

- Base UI drives the list as its composite root, so `scrollIntoViewIfNeeded`
  already runs against it: the active tab is revealed on mount and the focused
  one during arrow navigation, honouring `scroll-margin` on the tab and
  `scroll-padding` on the list. No JS of ours, no effect.
- `Tabs.Indicator` measures the active tab as `rect delta + list.scrollLeft`, and
  it is an absolutely positioned child of the list, so it scrolls with the tabs.
  The underline keeps tracking while the strip is scrolled.

Three supporting changes:

- `max-w-full` joins `w-fit`. `w-fit` alone cannot produce the scroll: the tabs
  are `shrink-0 whitespace-nowrap`, so the list's min-content width equals its
  max-content width and `fit-content` resolves to the tabs' full width whatever
  space it is given, leaving the strip overflowing its parent unscrolled. The cap
  is what turns that overflow into scrollable overflow; `w-fit` still hugs a
  strip that fits.
- `justify-center` becomes `justify-center-safe`. Plain centring centres the
  overflow too, putting the leading tabs at a negative offset outside the
  scrollable area, where no scroll position can reach them; `safe center` falls
  back to start alignment exactly when the strip overflows, so centred lists that
  fit are unchanged.
- The scrollbar stays hidden and containment is per axis. A bar would take block
  size from the strip the moment it appeared and paint over the indicator
  anchored to the strip's edge; the partially visible next tab carries the
  affordance instead. Containing overscroll on both axes would make a horizontal
  strip swallow the wheel gestures meant for the page, since it is a scroll
  container with no vertical range.

## Ink overflow

A scroll container clips ink overflow (box shadows, out-of-box paint) instead of
scrolling to it, so anything drawn outside a box near the strip's ends would lose
part of itself. Two paint sites move inside their boxes:

- `TabsTab` draws its focus ring inset. The inspector shell already did this
  locally for the same reason (its root's `overflow-hidden`); inset is the one
  placement no ancestor padding or scroll container can clip.
- The view switcher's drag-insertion line moves from the 2px gap beside the tab
  (`-start-0.5` / `-end-0.5`) to the tab's own edge (`start-0` / `end-0`).

## Measured behaviour

Static markup carries no layout, so the geometry was measured in Chromium on the
compiled classes, with a four-tab underline strip in a 240px parent:

| | without the cap | with `max-w-full` |
| --- | --- | --- |
| `clientWidth` / `scrollWidth` | 374 / 374 | 238 / 374 |
| overflows the parent | yes | no |
| scrollable | no | yes, 136px of range |

Focusing the last tab scrolls the list to 136 and brings it fully into view. The
indicator's painted left matches the active tab's at scroll offsets 0, 60 and
136, and its width matches the tab's, so the underline tracks under scroll. The
strip reports `overscroll-behavior-x: contain` with `overscroll-behavior-y:
auto`, and the hidden bar takes no block size (`offsetHeight - clientHeight` is
0). Plain `justify-content: center` puts the first tab at `offsetLeft: -55` in
the same fixture, which is the case `safe center` removes.

## Consumer audit

Eight call sites use `TabsList`. Seven pass text-only tabs and no width or
overflow assumptions (`guide-help-drawer`, `style-set-editor-controls`,
`clause-detail`, `memory-panel`, `billing-codes-dialog`, desktop settings,
`ui-playground`); none relies on clipping, and all render identically while their
tabs fit. The eighth is the view switcher, which wraps each tab in a positioned
`div` and paints the insertion line outside it, hence the change above.

The view switcher also carries the same affordance on a wrapper one level up
(`overflow-x-auto` plus hidden-scrollbar classes on the strip container, which is
also the drag drop target). That wrapper is now redundant for reachability but
still bounds the drag target, so it is left alone here rather than folded into
this diff; removing the duplicated scroll layer is a follow-up.

## Tests

`packages/ui/src/components/tabs.test.tsx` pins what static markup can carry: the
scroll container is the list element with the tabs and indicator inside it (a
nested scroll container between them fails), the width pair is `w-fit` with
`max-w-full`, centring is the safe variant, containment is axis-scoped, the bar
is hidden, consumer overrides still resolve through `tailwind-merge` (a merge
that stopped recognising `justify-center-safe` would silently pin every list to
centred), and every tab draws its ring inset. Reverting any of the primitive
changes fails these tests.

The merged inspector suite passes unchanged, including its focus-ring and
single-scroll-owner assertions.

### DIFF
--- a/.agents/skills/conventions-perf/SKILL.md
+++ b/.agents/skills/conventions-perf/SKILL.md
@@ -128,21 +128,39 @@ the route's `loader` either doesn't exist or never references `factory`.
 Prefetch `factory(...)` in the `loader` via `ensureRouteQueryData` or
 `prefetchRouteQuery`.
 
-## Depth-Jitter Caveat
+## How Depth Is Measured
 
-The chain-gap heuristic (`CHAIN_GAP_MS = 500` in `network.ts`) treats a
-request as "chained" onto the previous one only if it starts within 500ms of
-that request's end. Under load (a busy CI runner, a cold cache), independent
-requests that would normally fire in parallel can serialize past that gap by
-coincidence, and the computed depth flaps by 1 with **no change to the
-request manifest**. The checker carries a +1 depth allowance in
-`assertNetworkBaseline` for exactly this reason.
+`waterfallDepth` (`apps/web/e2e/helpers/network.ts`) counts the most
+consecutive **busy blocks** in any one observation sequence: a busy block is a
+maximal run of requests whose intervals overlap, and a launch gap over
+`REQUEST_SEQUENCE_GAP_MS` (500) starts a new sequence so an idle prefetch is
+not read as another route-load round. Two requests in flight at the same
+instant are never two levels, so the number is a lower bound on true causal
+depth: a dependent request hiding behind an unrelated slow one is invisible
+to it.
 
-If CI reports a deeper waterfall but the request list (`requests`,
-`requestCounts`) is unchanged from the baseline, treat it as scheduling
-jitter: rerun before touching the baseline. Do not bump an individual route's
-`depth` to paper over a flake; only `write`/`rewrite` when the request
-manifest itself actually changed.
+Reading launch times rather than the gap since the previous response *ended* is
+what buys monotonicity, and it costs sensitivity: a dependent request whose
+parent ran longer than 500ms opens a new sequence instead of counting a level.
+The three properties are not simultaneously satisfiable, because separating a
+dependent request from an idle prefetch requires the parent's duration, and any
+rule reading response ends lets a response growing under load close a gap it
+used to exceed. Under-counting is the safer error for a guard compared only
+upward.
+
+The metric is **load-monotone**: neither a slower response nor a uniformly
+stretched timeline can raise it. Sequence boundaries read launch times only,
+so a longer response cannot move one; coverage is a running max that is never
+rewound at a boundary, so a split can only lower the count.
+
+One residual remains, and it is what the +1 `DEPTH_JITTER_ALLOWANCE` in
+`assertNetworkBaseline` covers: two requests issued from different ticks may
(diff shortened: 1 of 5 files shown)


jan-kubica:
@codex review

jan-kubica:
Both review findings addressed.

**P1, dependent request after a slow parent.** Correct, and it is now a stated
trade-off rather than an oversight. `{start:0,end:600}` then `{start:600,end:610}`
reads 1, and a test pins it. The three properties are not simultaneously
satisfiable: telling a dependent request apart from an idle prefetch requires the
parent's duration, so any rule that reads response ends lets a response growing
under load close a gap it used to exceed, which is the non-monotonicity this PR
removes. Under-counting is the safer error for a guard whose baselines are only
compared upward: it costs a missed regression on a >500ms endpoint, where
counting ends costs recurring false failures on every route. The reasoning is in
the comment above `REQUEST_SEQUENCE_GAP_MS`.

**P2, driver timestamps for untimed requests.** Resolved by dropping the
`request.timing()` change entirely, so there is one clock again. The first CI run
on this branch measured what the switch actually does: browser timings report a
visibly shorter interval, blocks stop merging, and depth rises 1-2 levels on most
routes (`/chat` 3 -> 4, `/dev/autocomplete` 3 -> 5, `/contacts/$contactId` 4 -> 6).
That is a metric-scale change needing its own deliberate baseline re-derivation,
not something to land alongside a monotonicity fix. Recorded as a follow-up in
the PR description.

PR #2177, opened 2026-08-17
fix(api): normalize provider token counts in usage metering instead of panicking

jan-kubica:
## Proposed change

`computeRawUsageMicroUnits` (`apps/api/src/lib/usage/unit-model.ts`) is the shared boundary that turns provider-reported token usage into ledger micro-units. It treated an out-of-domain or inconsistent count as an internal invariant and called `panic()`, which aborts the whole process. Those counts are external input, so a single malformed or unusual payload can bring the process down:

- `cacheReadTokens` is only a subset of `inputTokens` by convention. A provider that reports more cache-read tokens than input tokens trips the `cache > input` panic, and also drives `billedInputTokens = inputTokens - cacheReadTokens` negative.
- A non-integer, negative, non-finite, or above-safe-integer value in any token field trips the non-negative-safe-integer panic.

Normalize at the boundary instead, mirroring the module's existing catalog-miss handling (capture telemetry + safe fallback rather than `panic`):

- Floor each token field into the non-negative safe-integer domain; out-of-domain values become `0`.
- Clamp `cacheReadTokens` to `inputTokens`, so billed input is never negative and the overlap bills at the cached rate.
- Report each anomaly once per model + field through `captureError` (deduped per process like the catalog-miss path, so a hot billing path can't spam telemetry).

Internal arithmetic-overflow guards stay as `panic()`; they protect our own rate arithmetic, not external counts.

## Checklist

- [x] **BUILD ASSURANCE** | Code builds correctly and without any errors or warnings. (`bun --filter @stll/api typecheck`, oxlint, and oxfmt are clean.)
- [x] **TESTING** | Added coverage for the cache clamp, per-field domain normalization, and the once-per-key telemetry. Mutation-checked: reverting the clamp or the floor fails the new tests.
- [ ] DARK MODE SUPPORT | N/A, backend metering change.
- [ ] ISSUE LINKED | No linked issue.
- [ ] SCREENSHOT INCLUDED | N/A, no UI change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Knc4tyLPR5symhSb9QHMWf)_

### DIFF
--- a/apps/api/src/lib/usage/unit-model.test.ts
+++ b/apps/api/src/lib/usage/unit-model.test.ts
@@ -137,18 +137,53 @@ describe("computeRawUsageMicroUnits", () => {
     ).toBe(0);
   });
 
-  test("panics on inconsistent token counts (cache > input)", () => {
-    expect(() =>
+  test("clamps cache reads above input tokens to the fully-cached result", () => {
+    const uncached = computeRawUsageMicroUnits({
+      modelId: "gemini-2.5-flash",
+      inputTokens: 100,
+      outputTokens: 0,
+    });
+    const fullyCached = computeRawUsageMicroUnits({
+      modelId: "gemini-2.5-flash",
+      inputTokens: 100,
+      outputTokens: 0,
+      cacheReadTokens: 100,
+    });
+    // Mutation guard: the cached rate is cheaper, so the clamp target is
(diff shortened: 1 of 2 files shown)


jan-kubica:
Superseded by #2180, which covers the normalization boundary plus the provider cache-accounting semantics and adapter canaries.

PR #2208, opened 2026-08-18
fix(chat): keep the streaming create-document preview mounted across source revisions

jan-kubica:
## Summary

While a `create-document` draft streams, the inspector recompiled the DOCX preview on every token and unmounted the `DocxEditor` whenever a partial revision failed to compile (a directive cut mid-token, a table row before its header row, an empty clause heading), then remounted it on the next token: a full parse, layout, and fit-width zoom snap per flash. The streaming branch also returned the bare editor while the ready branch wrapped it in `FileViewerWithAI`, so the end of streaming remounted it once more.

- `create-document-draft-preview.logic.ts`: `advanceCreateDocumentDraftPreview` keeps the last compiled document across streaming revisions the compiler rejects, recompiles only when the source or name changes, and shows no preview once a final source fails. The component adjusts this state during render (it depends on the previous render's result, not on the payload alone).
- Streaming revisions are coalesced with `useDebounce(source, 200ms, { leading, maxWait: 200ms })`; the final source applies immediately once streaming ends.
- The streaming draft renders inside the same `FileViewerWithAI` shell with the AI overlay deferred, so the editor keeps its tree position when the draft becomes ready; the review bar mounts only for ready/saving drafts (as before).
- Bumps folio (`@stll/folio-react` 0.14.0, `@stll/folio-core` 0.19.0, `@stll/docx-core` 0.14.0, `@stll/folio-agents` 0.8.6) and drops the deprecated `documentKey` prop at both call sites: folio now derives the external-load identity from its own load sequence (stella/folio#572), so a reload into an editor whose hidden view already exists (a click in a read-only preview creates one) replaces the content instead of freezing on the previous document. Without that, keeping the editor mounted across revisions would have exposed the freeze.

## Tests

`create-document-draft-preview.logic.test.ts` streams a production-shaped legal source one character at a time: asserts the stream passes through prefixes the compiler rejects (guards the carry-over against vacuity), the preview never drops after the first compiled prefix and only moves forward, one document is compiled per accepted prefix, a status change with the same source reuses the compiled document, and a failing final source shows no preview. Reverting the carry-over fails three of them.

## Summary by CodeRabbit

* **New Features**
  * Added smoother live previews while documents are being generated.
  * Previews now retain the last successfully compiled version during temporary streaming errors.
  * Preview updates are throttled to improve responsiveness.

* **Bug Fixes**
  * Improved handling of invalid or incomplete document content.
  * Failed final compilations now clear outdated previews.
  * Review controls are displayed only when a draft is ready or being saved.

* **Tests**
  * Added coverage for streaming previews, compilation failures, unchanged content, and empty documents.


jan-kubica:
@coderabbitai review

jan-kubica:
@coderabbitai review

jan-kubica:
@coderabbitai full review

PR #2298, opened 2026-08-20
perf: tighten TypeScript type boundaries

jan-kubica:
## Summary

- factor the shared optional permission shape out of the non-empty permission union
- preserve the structural requirement that every permission input names at least one resource
- measure TypeScript type and instantiation cost before retaining further boundary changes

This pull request remains draft while the isolated type-cost experiments are measured.


jan-kubica:
Closing while the experiments remain branch-only. A single consolidated pull request will be opened only after structural-equivalence checks and remote measurements justify every retained change.

CC on behalf of jan-kubica

PR #2309, opened 2026-08-20
fix: avoid duplicate statute reader payloads

jan-kubica:
- Keep workspace statute reads unchanged while public reads omit redundant full text when a usable document AST is available.
- Preserve full-text fallback for documents without a renderable AST.
- Cover structured and plain-text public reads with a database-backed invariant.

## Summary by CodeRabbit

- **New Features**
  - Public legislation reading now uses stored full text when structured document data is unavailable.
  - Structured legislation continues to prioritise parsed content and returns no unnecessary fallback text.

- **Bug Fixes**
  - Improved legislation text retrieval with a reliable fallback when the primary document structure cannot be used.

- **Tests**
  - Added coverage for full-text fallback behaviour across Czech and English legislation, including structured and unstructured documents.

### DIFF
--- a/apps/api/src/handlers/legislation/get.ts
+++ b/apps/api/src/handlers/legislation/get.ts
@@ -21,18 +21,24 @@ import {
 import type { EmptyAst } from "@/api/lib/legal-search/document-types";
 import type { LegislationReadDb } from "@/api/lib/legislation-public-read-db";
 
-type LegislationTextMode = "always" | "fallback";
+const LEGISLATION_TEXT_MODE = {
+  ALWAYS: "always",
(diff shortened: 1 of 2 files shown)


jan-kubica:
Implemented in `adb8c64f4d`: legislation text modes now use named values derived from one `as const` source. The focused public-read database suite passes.

CC on behalf of sok0

jan-kubica:
The named `LEGISLATION_TEXT_MODE` values and derived type are already present in d599ea4307; the defaults and comparison use those values. No further diff is needed.

CC on behalf of jan-kubica

PR #2320, opened 2026-08-21
chore: use Bun global store

jan-kubica:
## Summary

- enable Bun isolated installs backed by the shared global package store
- preserve project-graph resolution while declaring dependencies that isolation exposed
- keep Docker installs materialized because BuildKit cache mounts do not survive install layers
- allow Vite to serve package assets from the Bun store without broad filesystem access
- keep Expo Doctor strict while accepting only its known same-version Bun store false positive

## Why

Stella stayed on hoisted installs because [oven-sh/bun#23615](https://github.com/oven-sh/bun/issues/23615) made isolated catalog installs unreliable. Bun 1.4 fixed that blocker. The [global store](https://bun.com/docs/pm/global-store) lets local checkouts and Git worktrees share immutable package trees instead of storing a full copy in every checkout.

The public hoist pattern is migration compatibility for existing root scripts and workspaces; it is not the long-term dependency model. Remove it after each consumer declares every imported dependency. TypeScript preserves symlink paths so resolution follows the project link graph instead of package real paths inside the shared cache.

Three boundaries need explicit treatment. Docker disables the global store because its cache mount is detached after the install layer. Vite allows only the package-store subtree so linked fonts and assets remain servable. Expo Doctor currently treats Bun peer-context store entries as duplicate native modules ([expo#46429](https://github.com/expo/expo/issues/46429)); the wrapper accepts only same-version groups whose paths are confined to the project and Bun stores, and leaves every other Doctor failure fatal.

Strict isolation also exposed undeclared root tooling dependencies and the eslint-plugin-drizzle runtime import of @typescript-eslint/utils; both are declared explicitly.

## Validation

- frozen Bun install and workspace hygiene
- web and mobile typechecks; web production build
- Android, iOS, and web Expo exports
- real Expo Doctor run through the compatibility guard
- five compatibility-guard tests, including mutation verification
- Vite dependency canary against the local dev stack
- Docker contract tests and published-export guards
- type-aware lint on changed scripts; typecheck coverage; zero super-linear regex findings

CC on behalf of jan-kubica


jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

PR #2454, opened 2026-08-25
fix(api): grant corpus watermark retirement access

jan-kubica:
## Summary

- grant the ingestion role `DELETE` on case-law and legislation corpus delete-watermark tables through a forward migration
- prove the effective privilege with real `SET ROLE` deletes for both tables

## Context

Corpus generation retirement clears settlement watermarks after the corresponding search index is removed. The existing RLS policies admit that operation, but the table grants omitted the `DELETE` verb.

## Validation

- focused privilege test added; repository CI is the execution gate


jan-kubica:
@codex review

CC on behalf of jan-kubica

PR #2472, opened 2026-08-25
fix: repair corpus projection delete guard

jan-kubica:
## Summary

- replace the shared corpus projection delete guard with record-safe status access
- replay both the original and forward migration in the database contract test
- make timestamp precision explicit without lint suppressions

## Root cause

The delete trigger function is shared by intent and state history tables, but it referenced `OLD.status`; state rows do not have that field. The migration replay helper also skipped commented function DDL, so the test did not install the same trigger contract as production.

## Verification

- `bun run verify`
- corpus projection database contract: 7 tests, 100 assertions
- migration safety check
- suppression/type-cost ratchet
- mandatory pre-push lint, typecheck, formatting, and i18n checks

The regression test failed first on the missing trigger function, then on `record "old" has no field "status"`; it passes with the forward migration.

CC on behalf of jan-kubica

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to prevent removal of corpus projection history while related generations or projection processes remain active.
  * Improved validation of projection intent ordering and state transitions.
  * Ensured successful deletions return the affected record consistently.

* **Tests**
  * Updated database migration test coverage to validate the new deletion safeguards and related trigger behaviour.

### DIFF
--- a/apps/api/drizzle/20260825211300_corpus_projection_delete_guard_record/migration.sql
+++ b/apps/api/drizzle/20260825211300_corpus_projection_delete_guard_record/migration.sql
@@ -1,0 +1,37 @@
+SET lock_timeout = '1s';--> statement-breakpoint
+SET statement_timeout = '5s';--> statement-breakpoint
+
+CREATE OR REPLACE FUNCTION "guard_corpus_index_projection_history_delete"()
+RETURNS trigger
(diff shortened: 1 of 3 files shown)


jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

PR #2539, opened 2026-08-27
feat: adopt typed anonymization languages

jan-kubica:
## Summary

- update the native and WASM anonymization runtimes to 2.9.0
- type chat locale hints against the runtime-supported language set
- normalize regional locales and preserve the all-language name fallback for unsupported locales


jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

jan-kubica:
@coderabbitai review

CC on behalf of jan-kubica

PR #2561, opened 2026-08-28
fix: preserve Kanban drop axis types

jan-kubica:
## Summary

- preserve the closed group identifier union through Kanban schemas and resolved grouping
- carry the same union through saved workspace presentation and grouping picker callbacks
- return typed axis identifiers from atomic drop intents instead of widening them to `string`
- reject picker values that are not declared choices before crossing the typed callback boundary

## Validation

- `git diff --check`
- `bun test packages/ui/src/kanban/matrix.test.ts` (5 passing)
- workspace adapter test could not collect locally because the reused install does not contain `@dnd-kit/core`; CI will install the declared workspace graph

## Security

No new trust boundary, transport, persistence, authentication, or secret handling. The picker now narrows runtime select values against declared choices before calling the typed consumer.

## Summary by CodeRabbit

- **Improvements**
  - Kanban grouping identifiers are now preserved consistently across boards, saved views, grouping pickers and drag-and-drop actions.
  - Enhanced type support helps prevent mismatched grouping values when configuring or integrating Kanban views.
  - Grouping pickers now ignore unrecognised values while continuing to support clearing the grouping selection.
- **Bug Fixes**
  - Improved reliability when resolving active grouping properties and saved Kanban view state.

### DIFF
--- a/.changeset/typed-kanban-drop-axes.md
+++ b/.changeset/typed-kanban-drop-axes.md
@@ -1,0 +1,6 @@
+---
+"@stll/ui": patch
+"@stll/workspace-ui": patch
+---
+
+Preserve a board's group identifier union through grouping, saved-view, picker, and atomic drop-intent APIs.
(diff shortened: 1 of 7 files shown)


jan-kubica:
@codex review

CC on behalf of jan-kubica

PR #2580, opened 2026-08-28
fix: open every selected item in the inspector

### LINKED ISSUE #1671: bug(table): open all selected items in Inspector
### Reproduction

1. select multiple files in table view
2. click on "..." and click on Preview
3. only the first file opens

Image

### Expected behavior

In third step, all files open in the Inspector (ie right side viewer), focus is on the first one in the list

shanehobson:
## Proposed change

Closes #1671

In table view, Preview on a multi-row selection opened only the anchor row. Now every selected entity the inspector can render opens as a tab, in visible order, with the row the user acted on focused. From the toolbar menu (no anchor row) the first item in the selection is focused.

**Cause:** the Preview handler in row actions opened only the row the menu belonged to and ignored the rest of the selection. It now builds the list of tabs to open from every selected row and hands it to a new store action, `openTabs`, which adds all the tabs in one update and focuses the anchor.

The tree's context-menu Preview did loop over the selection, unlike the table, but with one store update per file, so it re-rendered per file and focus landed on whichever file came last. It uses `openTabs` now too and behaves the same.

One shared function (`planInspectorOpen`, built on `toFileTab`) decides which inspector tab an entity turns into: a task tab for tasks, a file tab for the first file the inspector can render, nothing for folders. Before, table row actions, the filesystem tree, kanban cards, and chat mentions each had their own copy of that logic, which is how the table and tree drifted apart. All four now call the shared function.

Side effect: if an entity's first file is something the inspector cannot render (a `.zip`, say) but a later file is a PDF, the PDF opens. The old copies looked at the first file only and gave up.

### Filesystem tree double-click (bundled, not in the issue)

Double-clicking a multi-selection in the tree also opened only one file, for a different reason. A double-click fires `click` first, then `dblclick`. The `click` handler selected only the clicked row, wiping the multi-selection, so by the time `dblclick` ran there was nothing bulk to open. A plain click (no Cmd/Ctrl/Shift) on a row already in a multi-selection now leaves the selection alone, as right-click already did. Modifier clicks still toggle or extend.

Selected tree rows also open in on-screen order rather than click order (the selection is a `Set`), so the tab strip matches the table.

## Checklist

- [x] DARK MODE SUPPORT | No visual changes.
- [x] ISSUE LINKED | Closes #1671
- [ ] SCREENSHOT INCLUDED | n/a

https://claude.ai/code/session_019QPKvNiLZFgDqYQ98G635s

## Summary by CodeRabbit

- **New Features**
  - Open multiple selected files and tasks together in the inspector.
  - Keep selected items in visible tree order while focusing the intended item.
  - Double-clicking a selected file opens the full selection.
  - Preview actions skip folders and unsupported files.
  - Converted files open using their available PDF version.
- **Improvements**
  - Unified opening behaviour across file rows, task cards, and row actions.
  - Existing inspector tabs are reused while preserving the requested focus.
  - File renames update immediately and remain consistent after saving.

### DIFF
--- a/apps/web/src/routes/_protected.workspaces/$workspaceId/-components/filesystem/tree-view.tsx
+++ b/apps/web/src/routes/_protected.workspaces/$workspaceId/-components/filesystem/tree-view.tsx
@@ -92,6 +92,7 @@ import { EmptyState } from "@/routes/_protected.workspaces/$workspaceId/-compone
 import { calculateFolderStatistics } from "@/routes/_protected.workspaces/$workspaceId/-components/filesystem/folder-statistics.logic";
 import type { FolderStatistics } from "@/routes/_protected.workspaces/$workspaceId/-components/filesystem/folder-statistics.logic";
 import {
+  getFileRowClickIntent,
   getFolderClickIntent,
   orderSelectedIds,
 } from "@/routes/_protected.workspaces/$workspaceId/-components/filesystem/tree-view-selection.logic";
@@ -1626,6 +1627,10 @@ const FilesystemRow = ({
     </>
   );
 
+  // `click` fires before `dblclick` and collapses a multi-row selection onto
+  // this row, so the opening click records the set it is about to collapse.
+  const dblClickSelectionRef = useRef<WorkspaceEntity[] | undefined>(undefined);
+
   // Double-click opens the whole selection, focusing the node clicked. The
   // context menu takes the same path through RowActions' `selectedEntities`.
   const openInInspector = () => {
@@ -1633,7 +1638,8 @@ const FilesystemRow = ({
       return;
     }
     openInspectorSelection({
-      entities: getBulkSelectedEntities() ?? [node],
+      entities: dblClickSelectionRef.current ??
+        getBulkSelectedEntities() ?? [node],
       anchor: node,
       workspaceId,
     })?.();
@@ -1674,15 +1680,27 @@ const FilesystemRow = ({
       className={rowButtonCls}
       onClick={(event) => {
         if (!isFolder) {
-          const meta = event.metaKey || event.ctrlKey;
-          // A plain click on a member of a multi-row selection keeps
-          // it, mirroring right-click, so the `click` before `dblclick`
-          // cannot collapse the selection the double-click opens.
-          if (isBulkSelected && !meta && !event.shiftKey) {
-            return;
+          const intent = getFileRowClickIntent({
+            clickCount: event.detail,
+            hasMeta: event.metaKey || event.ctrlKey,
+            hasShift: event.shiftKey,
+          });
+          switch (intent.type) {
+            case "select":
+              dblClickSelectionRef.current = intent.snapshotSelection
+                ? getBulkSelectedEntities()
+                : undefined;
+              onSelect(node.entityId, {
+                meta: intent.meta,
+                shift: intent.shift,
+              });
+              return;
+            case "keep-selection":
+              return;
+            default:
+              intent satisfies never;
+              return;
           }
-          onSelect(node.entityId, { meta, shift: event.shiftKey });
-          return;
         }
 
         // Shift extends a range like the file rows, taking priority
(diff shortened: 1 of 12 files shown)


jan-kubica (changes requested):
- [ ] **issue A** select rows 1 and 2, click Preview, both open (so far correct) but focus is on the second file. Nit, but IMO makes more sense to focus on the first selected? if this is intentional, feel free to keep as is

https://github.com/user-attachments/assets/add580e2-89f3-4a95-a643-d71e4cd95dad

- [ ] **issue B** sequence of clicking is this
1. click a file1
2. cmd(ctrl)+click another file, file2
3. plain click file1
4. BOTH file1 and file2 remain selected, BUG> only file1 should be selected now
5. shift+click file3
6. the range is file2-file3, proving the plain click did not update the selection anchor.

https://github.com/user-attachments/assets/df973f59-275c-4e51-a1a7-5ed3b958d459

apps/web/src/routes/_protected.workspaces/$workspaceId/-components/row-actions.tsx
```diff
@@ -370,35 +367,13 @@ export const RowActions = ({
...
-            workspaceId,
-          });
-      }
-      return undefined;
-    })();
+  // Preview opens every selected entity the inspector can render and
+  // focuses the row the menu was opened on.
+  const resolvedOnOpen = openInspectorSelection({
+    entities: bulkTargets,
+    anchor: entity,
```
jan-kubica, line 375: this is the issue A >  imo Bulk Preview should focus the first file in the ordered selection, not the row whose context menu was opened. With rows 1 and 2 selected, opening Preview from row 2 activates row 2 because it is passed as the anchor. Please derive the active tab from the selection’s first openable target.

However, the current UX in this PR is probably also defensible, I don't have a strong preference, if this is intended then please just add a short comment in code so it does not surprise users/ai in the future

apps/web/src/routes/_protected.workspaces/$workspaceId/-components/filesystem/tree-view.tsx
```diff
@@ -1718,10 +1674,14 @@ const FilesystemRow = ({
...
         if (!isFolder) {
-          onSelect(node.entityId, {
-            meta: event.metaKey || event.ctrlKey,
-            shift: event.shiftKey,
-          });
+          const meta = event.metaKey || event.ctrlKey;
+          // A plain click on a member of a multi-row selection keeps
+          // it, mirroring right-click, so the `click` before `dblclick`
+          // cannot collapse the selection the double-click opens.
+          if (isBulkSelected && !meta && !event.shiftKey) {
```
jan-kubica: this is the issue B >  > returning here prevents a plain click from reaching handleSelect, so the multi-selection does not collapse and the range anchor remains stale. 

please preserve double-click behavior without suppressing the completed single-click selection transition

apps/web/src/routes/_protected.workspaces/$workspaceId/-components/row-actions.tsx
shanehobson: Opening with the focus on the row whose context menu was opened was intentional - I do think it's a slightly better UX. But like you, I don't have a strong preference. I left a comment in the code explaining the reasoning behind the UX choice.

apps/web/src/routes/_protected.workspaces/$workspaceId/-components/filesystem/tree-view.tsx
shanehobson: Fixed - thanks for catching this!

https://github.com/user-attachments/assets/3bf00b0f-e0ad-4ebe-afaa-f21cab6520d8


https://github.com/user-attachments/assets/366ba188-8bd0-4366-b13f-7935c46fe9ac

shanehobson: One behavior change beyond what you reported, flagged so it is not a surprise: double clicking a single row opens the Preview and keeps the row selected. In `main`, doing this opens the Preview but deselects the row.

apps/web/src/routes/_protected.workspaces/$workspaceId/-components/row-actions.tsx
jan-kubica: OK, let's leave it as you have it, thank you for adding the comment

apps/web/src/routes/_protected.workspaces/$workspaceId/-components/filesystem/tree-view.tsx
jan-kubica: thank you for the boyscout fix:)

jan-kubica (approved):

PR #2581, opened 2026-08-29
feat: confirm bulk Preview above ten tabs

shanehobson:
## Proposed change

Stacked on #2580; the diff to review is the last commit (`feat: confirm bulk Preview above ten tabs`). Rebase onto `main` once #2580 merges.

While opening every selected item in the inspector (#2580), a large selection had no brake. User could select any number of files without limit. **Preview now asks first before opening files above a threshold** (similar to what bulk delete does).

- Above ten tabs the item opens a confirm dialog ("This opens N tabs at once.") with Cancel / Open. At or below ten it opens immediately, as before.
- New strings `workspaces.previewItems` and `workspaces.previewItemsDescription` added to every language file and `messages.gen.ts`.

## Checklist

- [ ] TESTING | Changes tested, including in different conditions (dark mode, language, narrow screen, etc).
- [x] DARK MODE SUPPORT | Uses the shared `AlertDialog`; no new styling.
- [ ] ISSUE LINKED | No issue; follow-up to #2580.
- [ ] SCREENSHOT INCLUDED | Pending manual check.

https://claude.ai/code/session_013CZ7hadAkH3oME4ykBYwdK

### DIFF
--- a/apps/web/src/components/chat/entity-open.ts
+++ b/apps/web/src/components/chat/entity-open.ts
@@ -5,6 +5,11 @@ import {
   isFileActiveInMainRoute,
 } from "@/components/chat/entity-route-detect";
 import { useInspectorTabsStore } from "@/components/inspector/inspector-tabs-store";
+import {
+  definedFields,
+  type FileTabSourceField,
+  toFileTab,
+} from "@/components/inspector/open-entities.logic";
 import { getTranslator } from "@/i18n/i18n-store";
 import { getAnalytics } from "@/lib/analytics/provider";
 import { api } from "@/lib/api";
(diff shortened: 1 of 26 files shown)


shanehobson:
Closing: the confirm isn't necessary UX. A bulk open is view-only and already reversible in one click via the tab context menu ("Close others" / "Close all"), and every other AlertDialog in the app guards a destructive or costly action. The perf concern is also already capped structurally: inspector-panel.tsx keeps at most MAX_MOUNTED_PDFS viewers mounted, so a large selection adds strip entries, not viewers.

Worth keeping from here, on #2580: the "Preview N items" menu label counting the tabs the plan will open, and routing the tree's double-click through the same open path. That surfaces the count without blocking the action.

CC on behalf of shanehobson

jan-kubica:
@shanehobson yeah, I tend to agree. I feel like for this and similar actions, its better to optimize for the usual setup, lets say in >95% of cases, user really wants to open all the tabs, so we should open them right away in order to make this experience smooth, but then make the potential mistake (<5% of cases) "cheap" as in allow to revert the action easily (in this case, there is a right-click > close all for the tabs or you could have a Toas with the button "close opened tabs").

PR #2591, opened 2026-08-29
fix: configure collaboration Valkey TLS

jan-kubica:
## Summary

- keep certificate verification enabled by default for collaboration `rediss://` connections
- allow an explicit deployment setting for private Valkey endpoints whose certificate has no trusted chain
- pass the policy to both ioredis clients created by the Hocuspocus Redis extension
- generate the collaboration environment example and cover secure defaults, explicit opt-out, and plaintext loopback behavior

## Validation

- `bun --filter @stll/collab typecheck`
- `bun --filter @stll/collab test src/env-schema.test.ts src/redis-options.test.ts src/server.test.ts` (35 passed, Redis integration skipped without its external test URL)
- `bun --filter @stll/collab format --check`
- type-aware oxlint on affected files
- `bun run env:check`

This is required by the infrastructure Valkey endpoint, which uses encrypted transport with a private self-signed certificate. The infrastructure opts out explicitly; every other deployment retains verification.

CC on behalf of jan-kubica

### DIFF
--- a/apps/collab/.env.example
+++ b/apps/collab/.env.example
@@ -22,3 +22,9 @@ STELLA_API_URL="http://localhost:3001"
 # [secret] Required. Bearer credential used by the collaboration service for
 # snapshot load and store requests.
 STELLA_COLLAB_SERVICE_TOKEN="local-collab-service-token-at-least-32-chars"
+
+# [internal] Optional with a default. Whether a rediss:// connection verifies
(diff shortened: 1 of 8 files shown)


jan-kubica:
@coderabbitai review
@codex review

Please review the exact head and the secure-default invariant.

CC on behalf of jan-kubica

PR #2601, opened 2026-08-29
fix: synchronize released workspace versions

jan-kubica:
Synchronize the three newly released workspace self-versions into `bun.lock`. This keeps `workspace:^` dependency ranges correct when Bun packs dependent packages.

The existing lockfile workspace-version guard passes, and dry-run tarballs resolve `@stll/ui` to `^0.11.0`.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -564,7 +564,7 @@
     },
     "packages/chat": {
       "name": "@stll/chat",
-      "version": "0.0.0",
+      "version": "0.1.0",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


jan-kubica:
Administrative bypass rationale: this lockfile-only repair intentionally skips hosted CI. The repository synchronizer and workspace-version guard pass, and fresh dry-run tarballs now resolve the released `@stll/ui` range. No additional workflow or review rerun was requested.

CC on behalf of jan-kubica

PR #2637, opened 2026-08-30
fix(web): preload inbox navigation data

jan-kubica:
Preloads the navigation query in parallel with the inbox feed so the route does not add a sequential request waterfall.

CC on behalf of jankubica


jan-kubica:
Trace evidence showed this prefetch was not the dependency creating the reported timing shape, so it is intentionally not being merged. CC on behalf of jankubica.

PR #2683, opened 2026-08-31
fix: preserve historical npm recovery

jan-kubica:
Allows an explicit one-package workflow dispatch to use the historical source without requiring the newly introduced selector script. This preserves the exact-artifact recovery path for a pre-existing immutable package tag.


jan-kubica:
@codex review

PR #2901, opened 2026-09-03
fix(landing): record the v0.8.13 release date

jan-kubica:
v0.8.13 was published and promoted; the landing changelog lists it with its date.


jan-kubica:
Folded into the v0.8.14 release PR, which records the prior release's date the way #2897 did.

PR #3017, opened 2026-09-06
fix(api): claim only scheduler jobs this build registers a task for

jan-kubica:
## Proposed change

`dueJobPredicate` selects a scheduler job on `enabled`, `nextRunAt` and the lease
columns, but not on whether the running build carries a handler for the row's
task. `runSchedulerOnce` therefore leases a job it may have no way to execute:
`runJob` looks the task up in the registry, finds nothing, and raises
`ConfigurationError`. The row spends a lease and a run record, is recorded as
failed, and is rescheduled to do the same thing on the next pass.

The registry is already the discriminator for this elsewhere: `registerSchedulerJobs`
retires a persisted row whose task no build declares any more, on exactly this
reasoning ("a persisted job whose task this build cannot run is undrivable"). It
cannot cover a row that is registered by another build and not by this one, which
is what a table shared across builds gives you.

- `acquireNextDueJob` takes the registry and filters the claim by its keys, read
  from the same instance the pass will look the handler up in, so the claim and
  the execution cannot disagree.
- The filter goes into the predicate rather than after the read. The claim takes a
  single candidate ordered by `nextRunAt`, so discarding an unrunnable row after
  the select would let the oldest one hold that slot on every pass and starve the
  jobs the build can run. A test pins that ordering property.
- With the claim restricted to registered tasks, a leased job without a handler is
  an internal invariant rather than a configuration a deployment can hold, so it
  panics instead of raising `ConfigurationError`.

## Checklist

- [x] **TESTING** | `bun --filter @stll/api test src/lib/scheduler/` (all suites pass); the two added cases fail against the unfiltered predicate. `bun --filter @stll/api typecheck`, `bun --filter @stll/api lint`.
- [ ] DARK MODE SUPPORT | Not applicable: no UI surface.
- [ ] ISSUE LINKED | No tracking issue.
- [ ] SCREENSHOT INCLUDED | Not applicable: no UI surface.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NGs5ALdf258KjeZc82N7n5)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Scheduler jobs are now claimed only when their tasks are available and runnable.
  - Unavailable tasks no longer block other eligible jobs from being processed.
  - Jobs that cannot be run remain unclaimed, preventing incorrect leasing and improving scheduler reliability.

### DIFF
--- a/apps/api/src/lib/scheduler/runner.test.ts
+++ b/apps/api/src/lib/scheduler/runner.test.ts
@@ -144,11 +144,13 @@ describe("acquireNextDueJob claim exclusivity", () => {
     const first = await acquireNextDueJob({
       db,
       leaseMs: LEASE_MS,
+      registry: noopRegistry,
       runnerId: "runner-a",
     });
     const second = await acquireNextDueJob({
       db,
       leaseMs: LEASE_MS,
+      registry: noopRegistry,
       runnerId: "runner-b",
     });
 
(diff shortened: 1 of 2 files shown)


jan-kubica:
Duplicate of #3016, which carries the same claim filter plus a sweep-level test and the empty-registry short-circuit. Closing in favour of that PR.

PR #3112, opened 2026-09-07
fix(api): read the NSS pagination state as a JavaScript literal

jan-kubica:
The NSS results page hands its pagination state to `infiniteScroll.js` in
single-quoted JavaScript literals, and the walk decodes them and posts the
value back to reach the pages after the first. Two things about that reader
were wrong. Both are invisible to the date-range search the crawl runs, and
both are fatal to any other one.

## The decoder

`unescapeJsStringLiteral` resolved `\uXXXX` wherever it appeared instead of
consuming escapes left to right, so the second half of an escaped backslash
started an escape of its own. Any condition whose value is itself quoted JSON
carries one: a condition drawn from a codelist states its options in
`ciselnikTreeData`, and each title there is wrapped in a backslash pair around
the escape for a double quote.

Decoding a captured codelist query with the old reader produces text that no
longer parses as JSON, which is then what the endpoint is asked for a page
with. The date-range condition has no nested quoting, which is why the simpler
reader stood.

The reader is now a left-to-right pass that consumes the backslash first and
handles the single-character escapes as well.

## The literal's boundary

`extractScriptString` matched `'[^']*'`, ending the literal at the first
apostrophe rather than the first unescaped one, so a value containing an
escaped apostrophe was posted truncated. It now matches `(?:\\.|[^'\\])*`.

## The empty 200

Both of the above make the endpoint answer 200 with an empty body, and so does
asking for a page past the day's last record. The walk could not tell them
apart and read either as the day being over.

That is case-law rule 14 from the wrong side. The crawl moved its cursor to the
next day, and being forward-only it never came back, so a refused query cost
every record past that day's first page, permanently and silently. The
signature is a day whose stored count stops at exactly 40.

`fetchResultPage` now refuses an empty body unless the day's stated record
count shows the page is past its last record, which is the one case where
emptiness is the answer. It throws `AdapterFetchError`, the same error the
function already throws for a non-ok status, with the day and page as its
cursor, and it invalidates the session as the other failure paths do. Throwing
holds the cursor, so the page is retried rather than skipped.

A day whose record count the page did not state is refused too: nothing
available at that point can show the day is over.

The reconciliation path already caught this indirectly, through the row-count
check that compares a page against the stated count. It now fails on the empty
body itself and names what it refused, rather than reporting a short page.

## Tests

Four new tests, each of which fails against the code before this change and
passes after it (verified by reverting the source alone):

- The codelist query is a verbatim capture from the portal, so the decoder is
  checked against the publisher's own bytes rather than against a restated
  expectation. The assertion that matters is not the byte string but that what
  gets posted still parses as JSON, since the endpoint reconstructs the search
  by parsing it.
- A literal ending at an escaped apostrophe is posted whole. This one is a
  constructed literal: the portal's own pages carry no apostrophe in these
  variables, so there are no real bytes to capture for it.
- An empty continuation body fails the crawl's page instead of ending the day.
- A page genuinely past the day's last record is still allowed to come back
  empty, so a day whose record count lands on a page boundary does not wedge.

The existing test that asserts the date-range query's posted byte string is
unchanged, and now says so explicitly: that path is byte-identical.

`bun scripts/run-tests.ts src/handlers/case-law/ingestion/adapters/cz-nss.test.ts
src/handlers/case-law/ingestion/adapters/cz-ns-reconciliation.test.ts` from
`apps/api`: 95 pass, 0 fail.

### DIFF
--- a/apps/api/src/handlers/case-law/ingestion/adapters/cz-nss.test.ts
+++ b/apps/api/src/handlers/case-law/ingestion/adapters/cz-nss.test.ts
@@ -47,6 +47,9 @@ const reconciliation = requireReconciliation(czNssAdapter);
 
 const BASE_URL = "https://vyhledavac.nssoud.cz";
 
+/** One backslash, for fixtures that must carry a literal escape. */
+const BACKSLASH = String.fromCodePoint(92);
+
 // ── Portal-shaped fixtures ───────────────────────────────
 
 /**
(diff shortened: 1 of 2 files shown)


jan-kubica:
Duplicate of #3111, opened by a retry of the same command. Continuing on #3111.

PR #3153, opened 2026-09-08
chore(ingestion): the deciding court comes from the record, enforced by lint

jan-kubica:
A publisher is not a court, and until now nothing said so. An adapter that
wrote its portal's own name into `court` attributed every decision that portal
republishes to a court that did not decide it, and `court` is what the
authority tier is read off, so the misattribution ranked too. The source-field
inventory did not catch it: a disposition names a target, not where the value
came from, so a constant satisfied a declaration that claimed to read the
page.

## The rule

`.agents/skills/conventions-ingestion/SKILL.md` gains rule 15: the deciding
court comes from the record, the court code in its ECLI first and then the
record's own court field, resolved through the shared resolver; never a
per-adapter constant, even where the portal is believed to serve one court's
decisions. An inventory declaration names the field the value is read from,
and a value produced from a literal satisfies no declaration.

## The guard

`no-literal-decision-court` reports a string or template literal written into
a `court` property anywhere under
`apps/api/src/handlers/case-law/ingestion/adapters/**` (adapter tests
excluded). The allowed form is a resolver call or a variable, which is what a
court read off the record looks like. `court` is the only court-valued key of
`IngestionResult`, so it is the only key watched; there is no `courtName`
field to guard.

- rule: `.oxlint-plugins/no-literal-decision-court.ts`, registered in
  `oxlint.config.ts` and catalogued in the plugin README
- fixture: `.oxlint-plugins/__fixtures__/no-literal-decision-court.fixture.ts`
  (rejected: the constant, an `as const` literal, a template literal, a parser
  argument; allowed: the resolver call, the source's own field, a shorthand)
- behavior test:
  `.oxlint-plugins/__tests__/no-literal-decision-court.test.ts` runs oxlint
  over both shapes and asserts which lines report. It runs from
  `scripts/lint-oxlint-fixtures.sh`, beside the fixture lint.

## Adapters

- `cz-us` resolves its court through the shared resolver, from the identifier
  the publisher states rather than the one `buildEcli` reconstructs: a
  reconstructed identifier carries this adapter's own court code, so a court
  read back from it would be the constant the call removes. The publisher's
  court stays as the fallback for a row that states nothing. A test asserts
  both paths store the court, so a dropped or bare-code attribution fails.
- `at-courts` renames its RIS `Gericht` query token to `courtFilter`. `AUSL`
  selects foreign-court rows; it is a listing filter, never an attribution,
  and naming it `court` was what made the guard ambiguous.
- `cz-ns` and `cz-nss` already resolve through the shared resolver.

No suppression was added to make an adapter pass.

Adapters outside the Czech sources take their court from the record's own
field already (`sk-courts` from the item's court, `pl-courts` from the item's
court name, `at-*` from the RIS metadata, `eu-ecj` from the ECLI). `sk-us`
assigns the publisher's court to a local variable, which the rule allows: its
source has no court field and there is no Slovak ECLI court map to resolve
against, so routing it would mean inventing one. It is the remaining place
where a single-court assumption is only an assumption.


jan-kubica:
Stacked on #3151: it introduces the shared court resolver this PR routes `cz-us` through and enforces with lint. Base retargets to main once #3151 lands.

PR #3263, opened 2026-09-12
fix: add inspector chat scroll padding so actions clear follow-ups

### LINKED ISSUE #1684: bug: allow for more scroll in right-side Inspector chat
### Reproduction

See screenshot, the Copy and Retry buttons are partially overlapping with the next suggested question(s) despite user scrolling to the very bottom of the chat.

Make it so there is a bit of artifical extra space allowing user to scroll a bit more, in order to avoid the overlap of buttons with the suggested questions

Image

### Expected behavior

no overlap

minutechreview:
### Description

When Inspector follow-up suggestion chips are showing, the transcript now uses a bit more bottom padding (`pb-44` instead of `pb-32`) so Copy/Retry can scroll above the chip row. Padding stays `pb-32` when there are no chips.

### Motivation

Fixes #1684 — Copy and Retry overlapped the next suggested questions even when scrolled to the bottom.

### Additional details

- File: `apps/web/src/components/inspector/chat-tab-panel.tsx`
- Spacing-only change using existing Tailwind padding classes
- Before/after screenshots: not included (could not run the app in the agent environment). Happy to add if maintainers want a follow-up

Fixes #1684

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat transcript spacing when suggested follow-up chips are displayed, keeping Copy and Retry actions accessible without being obscured.

### DIFF
--- a/apps/web/src/components/inspector/chat-tab-panel.tsx
+++ b/apps/web/src/components/inspector/chat-tab-panel.tsx
@@ -625,8 +625,12 @@ export const ChatTabPanel = ({
         >
           <Conversation className="min-h-0 flex-1">
             {/* Bottom padding keeps the last messages readable above the
-                floating composer block (veil + pill + status row). */}
-            <ConversationContent className="gap-3 pb-32">
+                floating composer block (veil + pill + status row). Extra
+                space when follow-up chips are showing so Copy/Retry can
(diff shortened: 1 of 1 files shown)


jan-kubica:
thank you!

PR #3318, opened 2026-09-13
fix(case-law): tell the refine box apart from find-in-table

jan-kubica:
The public results toolbar drew two magnifiers side by side: the find-in-table trigger and the refine box, which the shared Input decorates on `type="search"`.

- The refine box now takes a funnel through `InputGroup`/`InputGroupAddon`, the sanctioned adorned input; label, placeholder and tooltip are unchanged.
- Its `h-7 min-h-0 text-xs` override landed on the Input's control while the field kept the shared metrics, so the absolutely positioned glyph centred on 28px and the text on 30px. The override is gone; the field uses `size="sm"` like the find bar's own input.
- `INPUT_CONTROL_CLASS_NAME` now centres its field, so a caller height on the control can no longer desync the glyph from the text in any adorned input.

Find-in-table itself is untouched: `TableFindBar` was already the one component behind both toolbars.


jan-kubica:
Superseded by #3319, which removes the refine box this changed.

PR #3455, opened 2026-09-16
fix(auth): lenient and roomier dynamic client registration

jan-kubica:
Dynamic client registration rejected a body with `contacts: []`, which a real MCP client sends. RFC 7591 §2 reads an empty optional list as absent, so the OAuth provider patch drops the non-empty constraint on `contacts` and normalises `[]` to absent; every other registration check is unchanged. Unknown top-level metadata was already stripped and is now pinned by a fixture.

`apps/api/src/tests/helpers/oauth-client-registration-fixtures.ts` holds the registration bodies real MCP clients send (captured 2026-09-16, volatile values replaced), the repository's own two bodies, synthetic leniency cases, and two negative controls. `apps/api/src/lib/oauth-client-registration.db.test.ts` posts each one to the real register endpoint over PGlite. Reverting the patch turns exactly the two `contacts` fixtures red with the original error. A further fixture encodes Microsoft's documented registrar constraints (secret required, PKCE, Microsoft redirect hosts) until a capture replaces it.

The register endpoint's rate limit rises from the library default of 5 to 30 per minute per address (`AUTH_RATE_LIMITS.oauthClientRegistration`): clients register once per user connect, a team behind one NAT connects within a minute, and a test pins the constant to the rule the plugin enforces.


jan-kubica:
The first finding is addressed in cde18ff4c9: `buildOAuthClientRegistrationRequest` now owns the request-body construction, `registerOAuthClient` posts its output, and the census fixture is built from that builder with test inputs, so it can no longer carry a stale copy. The second is not landed as suggested: declaring the provider patch as a `@stll/api#test` input fails `check:test-input-coverage` ("which no test in apps/api reads any more"), because the suite depends on the patch through the installed package rather than by reading the file, so binding a patch-only edit to this suite needs a repo-wide input such as `globalDependencies: ["patches/**"]` rather than a per-task one.

jan-kubica:
Finding 2 is now closed at the right layer: `patches/**` is a Turbo global dependency (f447a1ee3e), so a patch-only edit invalidates every task rather than one suite's inputs.

PR #3561, opened 2026-09-18
chore(lint): guard motion and viewport rules from the UX conventions

jan-kubica:
- Stylelint: `transition`/`transition-property` may not be `all` or name a layout property; `@keyframes` frames may not animate one.
- oxlint `no-layout-motion-classes`: reports `transition-all`, `transition-[…]`/`animate-[…]` naming layout properties, and `h-screen`/`w-screen` family classes; autofix to `transition` and `dvh`/`dvw`.
- 15 existing `transition-[…]` sites in the design system (sidebar rail, accordion panel, popover positioners, usage bar, alphabet rail) animate box properties by design; they are listed in `oxlint.config.ts` as a shrink-only allowance.
- The UX skill names the enforcing rules on the three bullets now guarded.


jan-kubica:
Self-review found a conflict between the documented transform/opacity-only convention and 234 existing paint-transition call sites. This update keeps the documented convention: layout motion and static viewport utilities fail directly, box-owning layout exceptions are exact file-and-utility pairs, and existing paint transitions in web, desktop, landing, UI, and CSS are recorded per file in a decrease-only ratchet. The follow-up cleanup is to remove those paint transitions and rewrite the baseline downward; new call sites cannot increase it.

CC on behalf of jan-kubica

PR #3595, opened 2026-09-19
fix(api): bound the corpus search vector input to the tsvector ceiling

jan-kubica:
## Proposed change

`indexLegislationDocument` and `indexDecision` compute their `tsv` column as
`to_tsvector(<regconfig>, unaccent(arabic_normalize(title || ' ' || searchable_text)))`
over the document text as read. Postgres builds a tsvector's lexeme buffer against a
fixed 1048575-byte ceiling and raises SQLSTATE 54000 from `make_tsvector` above it, so a
long enough statute or decision makes that upsert a statement it cannot satisfy. Such a
document is not partially indexed: the row never lands, the backfill probe reads the
document back as missing on the next pass, and the pass fails on it the same way, for
good.

Both call sites now build the expression through `corpusSearchVector`, which bounds the
normalized text with `left(...)` to `LIMITS.corpusSearchVectorMaxChars`. The bound wraps
the normalization rather than the columns, because unaccenting expands some characters
into several, so only a bound over the text `to_tsvector` actually reads holds. Past the
bound the tail stops being searchable, which is what the document trades for being
indexed at all.

The bound: a character costs at most its four UTF-8 bytes of lexeme plus the four-byte
entry its token amortizes, so a fifth of the ceiling stays under it whatever the script.
Measured against Postgres, distinct five-letter ASCII words cost 1.67 bytes per
character and distinct five-letter Cyrillic words 2.33.

`corpus-search-vector.db.test.ts` builds a document whose distinct lexemes overrun the
buffer, asserts Postgres rejects it unbounded (SQLSTATE 54000) so the fixture reaches
the fault, then asserts the bounded expression indexes the head of the same text and
drops the tail. Removing `left(...)` turns the test red.

## Checklist

- [ ] DARK MODE SUPPORT | User can use the webapp in dark mode. Make sure your changes are visible in both light and dark mode.
- [ ] ISSUE LINKED | Link an issue to this PR (not by mentioning it in the description, but by using the GitHub issue linking feature)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xwqbf23j1mQ66mmb58GxTc)_


jan-kubica:
Closing in favour of #3594, which bounds the text only after Postgres refuses the projection. An up-front bound at a fifth of the ceiling drops the tail of every long statute from search, while the ceiling is on the vector rather than the text: ordinary prose dedupes into far fewer bytes than it occupies, so most long documents fit whole. #3594 now covers the case-law projection as well.

PR #3601, opened 2026-09-20
fix(case-law): match ECLI court codes that carry a court number

jan-kubica:
## Proposed change

The ECLI pattern in `citation-extractor.ts` typed the court code as letters-only:

```
/ECLI:[A-Z]{2}:[A-Z]{1,8}:\d{4}:[\w.]+/gu
```

Per the ECLI specification the court code is alphanumeric, and Slovak cities holding
several courts of one kind number the duplicates, with the number part of the code
rather than a suffix on it. The letters-only class therefore matches no numbered court
at all, silently: `ECLI:SK:OSKE1:2018:7117220342.4` (Okresný súd Košice I) and
`ECLI:SK:OSKE3:2014:7513207505.1` (Okresný súd Košice-okolie) yield no citation,
while the letters-only `ECLI:SK:NSSR:2019:…` on the same input matches.

This widens the class to alphanumeric. The code still has to open with a letter, so a
missing code cannot let the year stand in for one, and the quantifier stays bounded,
keeping the scan linear.

Both spellings above were seen in production corpus text. The fixture in the test is
verbatim decision-header prose, with the judge field dropped.

Note on `pipeline.ts`: a decision's own ECLI is filtered downstream by
`isSelfCitation`, so nothing here makes a decision cite itself. Seeing the identifier
is the point, because that spelling is what *other* decisions cite it by.

## Checklist

- [x] **TESTING** | `bun --smol test src/handlers/case-law/ingestion/citation-extractor.test.ts` passes (181 tests). The added test fails on the pre-change pattern and passes after, so it guards the behavior rather than restating it.
- [ ] BUILD ASSURANCE | Not run locally; only `oxfmt` and the extractor suite above were run against the two touched files. Deferring to CI.
- [ ] DARK MODE SUPPORT | N/A, no UI surface.
- [ ] SCREENSHOT INCLUDED | N/A, no UI surface.
- [ ] ISSUE LINKED | No tracking issue.


jan-kubica:
The merge queue dropped this PR on a CI failure that does not come from this change.

**Failing check:** `CI Checks / e2e-production-shard (1)`, queue run 17247 (`gh-readonly-queue/main/pr-3601-798c98bc`).

```
[chromium] › e2e/specs/evidence-references.spec.ts:157:5 › DOCX evidence references (en)
  › persists references, resolves the source, and renumbers after reload
Error: expect(locator).toBeVisible() failed
Locator: locator('[aria-label="Document content"][contenteditable="true"]')
Timeout: 45000ms — element(s) not found
  at apps/web/e2e/specs/evidence-references.spec.ts:312:9
```

Why it is not this PR's:

- The same head commit (`844b674`) passed the full merge-queue CI in run 17245 nine minutes earlier, on a different queue base. Same tree, opposite result.
- In the failing job the same spec passed for two other locales (`cs` 10.9s, `ar` 7.4s); only `en` failed, at 52.6s against a 45s visibility wait. That is a timing boundary, not a deterministic break.
- The diff is a character class in `apps/api/src/handlers/case-law/ingestion/citation-extractor.ts` plus its test. Nothing in it reaches the editor surface that assertion waits on, and widening a citation regex cannot stop a contenteditable from mounting.

No existing fix to port. I have deliberately not touched `evidence-references.spec.ts`: making that wait robust belongs in `apps/web` as its own change, and folding it in here would widen a two-file fix into an unrelated subsystem.

The PR is green and mergeable on its own head, and it was already re-queued at 08:01, so no action is needed on it here.

CC on behalf of jan-kubica

PR #3616, opened 2026-09-20
fix(api): admit skill catalog tools when validating a chat continuation

shanehobson:
I ran into a bug when implementing the `save_playbook` tool in https://github.com/stella/stella/pull/3614:

## Proposed change

In an organization with an installed skill, answering an `ask-user` question (or an approval) fails with HTTP 400 when the model called `load-skill` or `read-skill-resource` earlier in the same turn. The client shows "The request could not be completed", and resending the answer fails the same way, because the continuation carries the same assistant message.

### Cause

`send-message.ts` builds two tool sets per request with `getChatTools`:

- The **run** set, what the model may call this turn. It receives `skillMetadata`, the organization's installed skills plus the built-ins, so it registers `load-skill`.
- The **validation** set, which `validateMessage` checks the incoming message's tool calls against (the tool must exist, and its `arguments`, `input`, and `output` must parse). It never executes. It was built without `skillMetadata`, so `getChatTools` fell back to `getChatSkillMetadata()`, the built-in skills. That list is empty, and `createSkillTools` registers no catalog tool for an empty list.

On a continuation, the tool calls come from the persisted assistant message. So a call the server itself authored was rejected with `Unknown chat tool: load-skill`.

### Timeline

- **#300** (user-managed skills) made the run set read the organization's catalog. The validation set was not given the same input. From then on the two sets could disagree about skills.
- Until **#1058**, four built-in skills shipped under `packages/skills/skills/`. The validation fallback was therefore never empty and always registered `load-skill`, which hid the mismatch.
- **#1058** removed the four built-in skills (`GENERATED_SKILLS` is `[]`). Since then, any organization with an installed skill can reach the 400.

It stayed unnoticed because three things must coincide in one turn: an installed skill exists, the model loads it, and the model then pauses for an answer or an approval. Only the incoming message is validated, not the thread's history, so a turn that loads a skill and runs to completion never trips it. No test sent a continuation whose message held a skill tool call.

A second route to the same 400 is latent. For a catalog holding only built-in skills, `skillName` is an exact `picklist`. Once a built-in skill ships again, a persisted call naming a skill that was later renamed or removed would fail validation in an organization with no installed skills.

### Fix

Validation proves that a message's calls are well-formed calls to tools stella defines. Whether a skill is available is decided when the tool runs (`assertAvailableSkill`, `loadAvailableChatSkill`), inside the caller's organization, and a validation set never runs. So in a validation set, `createSkillTools` registers `load-skill` and `read-skill-resource` whatever the catalog holds, with `skillName` as a plain string, the same schema a run with installed skills already uses. Both tools keep a single definition; only the catalog-dependent part differs. The run set is unchanged: an empty catalog still exposes no `load-skill`, and a built-in-only catalog keeps its `picklist`.

### Known gaps, listed and not fixed

The guard found more of the same class. The validation set honors three gates that are read from organization or thread state on every request: the thread's web-search setting, the resolved web-search provider, and `disabledNativeToolSlugs`. If one closes while a turn awaits an answer or an approval, the continuation's persisted calls to the gated tools (`web_search`, `fetch_url`, the `boe` and `infosoud` tools) no longer validate.

This is also a documentation conflict. The `disabledNativeToolSlugs` prop comment in `chat-tools.ts` says "Validation tool sets ignore this"; `send-message.ts` passes it to the validation call and says validation tools "still honor thread/org gates". Deciding which is right reverses a documented choice, so it is left for a separate change. The guard pins the three gaps in a test with derived expectations, so they can only shrink. Proposed unification: in a validation set, ignore `disabledNativeToolSlugs` (one line per group), and give the web-search tools a schema-only construction path that needs no provider.

## Checklist

- [N/A] DARK MODE SUPPORT | User can use the webapp in dark mode. Make sure your changes are visible in both light and dark mode.
- [N/A] ISSUE LINKED | Link an issue to this PR (not by mentioning it in the description, but by using the GitHub issue linking feature)
- [N/A] SCREENSHOT INCLUDED | If relevant, please include a screenshot / video of the functionality added (allows for a quick check of minor ux changes)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved continuation of chat turns paused while awaiting user input after skill-tool calls.
  - Chat validation now recognises available skill names and accurately reflects configured tools without executing them.
  - Invalid or unknown tool names continue to return a clear 400 error without starting a stream.

- **Tests**
  - Added comprehensive coverage for skill, document-editing, web-search, memory, role, and mode configurations.


jan-kubica (dismissed):

jan-kubica:
thank you, this is correct. 
i have pushed one extra commit that moves the fix a level down to address the other places you have listed in PR that are affected by the same bug shape

